PR #256, opened 2016-06-02
Update README.md

wardweistra:
Typo in MPAndroidChart


mikecarroll:
Good catch! Though, just to be fair, everybody loves chard...

PR #275, opened 2016-09-02
adding back the task view change

dariosalviwork:


dariosalviwork:
I've simplified a bit the interaction between the ViewTaskActivity and the Task class.
It's not perfect, but it doesn't disrupt the rest of the code.

mikecarroll:
@dariosalviwork Awesome. @nrlakin Can you take a quick look with regard to the issues you were experiencing before?

dariosalviwork:
hello,
I've added the possibility of generating the password instead of asking it to the user. This also excludes the possibility to re-join the study without having to re-register.

backbone/src/main/java/org/researchstack/backbone/model/DocumentProperties.java
```diff
@@ -88,4 +90,24 @@ public void setRequiresBirthdate(boolean requiresBirthdate)
...
+    {
+        return generatePassword;
+    }
+
+    public void setGeneratePassword(boolean generate)
+    {
+        this.generatePassword = generate;
+    }
+
+    public int passwordLenght()
```
liujoshua: typo in Length

backbone/src/main/java/org/researchstack/backbone/storage/database/sqlite/DatabaseHelper.java
```diff
@@ -82,7 +82,7 @@ public void saveTaskResult(TaskResult taskResult)
             taskRecord.completed = taskResult.getEndDate();
             getDao(TaskRecord.class).create(taskRecord);
 
-            Gson gson = new GsonBuilder().setDateFormat(FormatHelper.DATE_FORMAT_ISO_8601).create();
```
liujoshua, line 85: This looks like a backwards incompatible change, is it needed to fix a bug?

backbone/src/main/java/org/researchstack/backbone/task/Task.java
```diff
@@ -40,6 +46,53 @@ public Task(String identifier)
...
+    public ViewTaskActivity getActivity(){
+        return this.activity;
+    }
+
+    /**
+     * Forces the task to proceed to a given @{@link Step}.
+     * Results, if any, should be computed before.
+     * @param step the @{@link Step}
+     */
+    public void forceStep(@NonNull Step step){
```
liujoshua, line 77: How does this work with getStepAfterStep and getStepBeforeStep?

backbone/src/main/java/org/researchstack/backbone/task/Task.java
```diff
@@ -40,6 +46,53 @@ public Task(String identifier)
     }
 
     /**
+     * Sets the current @{@link ViewTaskActivity} for this task.
+     * This function should be called only by the {@link ViewTaskActivity} instance running
+     * at its creation time.
+     * @param activity the @{@link ViewTaskActivity} instance
+     */
+    public void setActivity(ViewTaskActivity activity){
```
liujoshua, line 54: I prefer the way onViewChange was handled by Activity, What is the intent of this change? 

setActivity, getActivity, forceStep, and forceFinish are written in a way that exposes ViewTaskActivity to Task, which doesn't seem ideal for several reasons

- Tasks now can hold a reference to activities, which are created/destroyed by Android, and would have to be dereferenced properly. It's best to avoid holding a reference if it can be avoided. 
- Tasks now have to know about activities, which seems like an inversion of their respective responsibilities. The activity already knows about the Tasks and Steps. Here, we have ViewTaskActivity calling something on a Task, and Task registers a callback with activity. To me it would make more sense to have Activity ask Task for a callback, and register it.

backbone/src/main/java/org/researchstack/backbone/ui/callbacks/StepViewCallback.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+public interface StepViewCallback {
+
+    /**
+     * This function is called when the {@link org.researchstack.backbone.ui.step.layout.StepLayout}
+     * of a @{@link Step} has been visualised and all animations have finished.
+     * @param activity the instance of the {@link ViewTaskActivity}
+     * @param step the @{@link Step} currently being shown
+     */
+    public void onStepShown(ViewTaskActivity activity, Step step);
```
liujoshua, line 18: This could be an inner class in ViewTaskActivity, since it should only be used with ViewTaskActivity.

Looks like onStepShown is intended to be called from the UI thread, so I suggest adding @MainThread annotation.

ViewChangeType seems important, I see you originally wrote the code which added that. Why did you remove it?

backbone/src/main/java/org/researchstack/backbone/ui/views/StepSwitcher.java
```diff
@@ -98,8 +99,9 @@ private void init()
      * @param stepLayout the step you want to switch to
      * @param direction  the direction of the animation in the x direction. This values can either be
      *                   {@link StepSwitcher#SHIFT_LEFT} or {@link StepSwitcher#SHIFT_RIGHT}
+     * @param finished a piece of code that is executed when the animation is completed (can be null)
      */
-    public void show(StepLayout stepLayout, int direction)
+    public void show(StepLayout stepLayout, int direction, @Nullable Runnable finished)
```
liujoshua, line 104: Not sure if this change is still needed, if we go back to having the activity handle notifying the Task of the step change. But if there is a desire for this, let's keep the old method signature and have it call this one (passing null for finished) to maintain backwards compatibility.

skin/src/main/java/org/researchstack/skin/DataProvider.java
```diff
@@ -105,7 +106,7 @@ public static void init(DataProvider instance)
      * @return Observable of the result of the method, with {@link DataResponse#isSuccess()}
      * returning true if signIn was successful
      */
-    public abstract Observable<DataResponse> resendEmailVerification(Context context, String email);
```
liujoshua, line 108: Why should password be required to resend an email verification? Usually email verification is done during registration, and should be the same independent of whether the user has changed their password, has created an account (I could imagine verifying email prior to sign up).

Secondly, this is a backwards incompatible change, suggest adding back the method with the old signature.

skin/src/main/java/org/researchstack/skin/model/User.java
```diff
@@ -7,26 +7,60 @@
...
+     */
     public void setName(String name)
     {
         this.name = name;
     }
 
+    /**
+     * Gets the password of the user, as registered on the server.
+     */
+    public String getPassword()
```
liujoshua, line 50: How are you planning to use this class? It's hard to tell because it's not used in this pull request. User isn't even used internally by ResearchStack, so I'd argue it should be deprecated from RS and the implementation left to the app.

In Mole Mapper, the User class is used to serialize/deserialize information used during registration and consent, to remember who the user is, prior to consent/registration completion, e.g., it's used to repopulate the user's email in the sign in screen, while leaving the password empty.

Information like password is stored via other mechanisms.

skin/src/main/java/org/researchstack/skin/task/ConsentTask.java
```diff
@@ -38,15 +38,13 @@
...
 {
-    public static final String ID_VISUAL                        = "ID_VISUAL";
     public static final String ID_FIRST_QUESTION                = "question_1";
     public static final String ID_QUIZ_RESULT                   = "ID_QUIZ_RESULT";
     public static final String ID_SHARING                       = "ID_SHARING";
     public static final String ID_CONSENT_DOC                   = "consent_review_doc";
     public static final String ID_FORM                          = "ID_FORM";
     public static final String ID_FORM_NAME                     = "ID_FORM_NAME";
     public static final String ID_FORM_DOB                      = "ID_FORM_DOB";
-    public static final String ID_FORM_BIRTHDATE                = "ID_FORM_BIRTHDATE";
```
liujoshua, line 49: Why remove these? This is a backwards incompatible change

skin/src/main/res/values/strings.xml
```diff
@@ -139,7 +139,10 @@
     <string name="rss_local_error_load_task">Unable to load the selected task</string>
     <string name="rss_error_select_answer">Please select an answer</string>
     <string name="rss_error_invalid_password">Invalid Password</string>
+    <string name="rss_error_small_password">%1$d characters minimum</string>
     <string name="rss_error_invalid_email">Invalid Email</string>
+    <string name="rss_repeat_signup">Your account has been signed out, you need to repeat the sign up.</string>
```
liujoshua, line 144: Are we talking about sign out and sign in? Or deregister and sign up?

skin/src/main/java/org/researchstack/skin/ui/SplashActivity.java
```diff
@@ -30,14 +30,14 @@ public void onDataReady()
                 .initialize(this)
                 .compose(ObservableUtils.applyDefault())
                 .subscribe(response -> {
+                    boolean boarded = AppPrefs.getInstance(this).isOnboardingComplete();
+                    boolean signedin = DataProvider.getInstance().isSignedIn(this);
+                    boolean emailverified = AppPrefs.getInstance(this).isEmailVerified();
 
-                    if(AppPrefs.getInstance(this).isOnboardingComplete() ||
```
liujoshua, line 34: This is a significant behavior change. Looking at the documentation for setOnboardingComplete in AppPref, this behavior would be unexpected.

Also, I see that OnboardingActivity already calls EmailVerificationActivity. So it would make sense to me the OnboardComplete flag in AppPrefs could handle it.

skin/src/main/java/org/researchstack/skin/schedule/ScheduleHelper.java
```diff
@@ -14,15 +18,37 @@
...
+    /**
+     *
+     * @param cronString the cron string that describes the periodicity
+     * @param delay delay in ISO 8601 since the scheduling was actually started the first time (after onboarding)
+     * @param interval minimal amount of time between two executions, complements the chron information
+     * @param usestart when the app was started first, should be after onboarding and never null
+     * @param lastExecution last time the task was completed, can be null
+     * @return
+     */
+    public static Date nextSchedule(String cronString, String delay, String interval, @NonNull Date usestart, Date lastExecution)
```
dephillipsmichael, line 30: ScheduleHelper is going to soon be deprecated.  See https://sagebionetworks.jira.com/browse/MA-35 or it is a story describing steps to make an ActivitiesManager similar to what is in iOS.

backbone/src/main/java/org/researchstack/backbone/storage/database/sqlite/DatabaseHelper.java
dariosalvi78: the actual format used here is not ISO 8601, it's RFC 822. Writing ISO 8601,as originally done, is a mistake and is misleading.

I have tried to implement the real ISO 8601 (which is used by ResearchKit BTW), but it's not well supported in Android unfrotunately and I had to revert back to this.

backbone/src/main/java/org/researchstack/backbone/task/Task.java
dariosalvi78: The current desing fits very well with tasks, like surveys, where the flow is controlled by the user, but fits extremely poorly with tasks where the flow is controlled in a different way.
These changes were necessary to implement the "timed walk" task. I needed to access the activity as the task was determining what was being displayed and not the activity imposing the when to go "next". Also, my task needs to have access to the context in order to retrieve data from sensors and other amenities. I couldn't find any way to implement these features without having to break the current design completely (as done in MoleMapper for example). I know it's not ideal but it was the best solution I could find.

backbone/src/main/java/org/researchstack/backbone/task/Task.java
dariosalvi78: getStepAfter/Before is called by the activity (when, e.g. a button is pressed).
forceStep() is "mandated" by the task.

backbone/src/main/java/org/researchstack/backbone/ui/callbacks/StepViewCallback.java
dariosalvi78: There was a separate callback used to tell when the view task activity was create, paused, etc. But that is not needed any more if the Task has complete access to the activity instance.

backbone/src/main/java/org/researchstack/backbone/ui/views/StepSwitcher.java
dariosalvi78: sure...

skin/src/main/java/org/researchstack/skin/DataProvider.java
dariosalvi78: that's how it works in my application. I am using a server which is partially compatible with Sage and the email needs the password too. I can leave the old signature too.

skin/src/main/java/org/researchstack/skin/model/User.java
dariosalvi78: I am using it as it is being used in MoleMapper. But I need to keep more information than the one used in Mole Mapper. I am OK to remove this class from RS and leave it to the app.

skin/src/main/java/org/researchstack/skin/task/ConsentTask.java
dariosalvi78: not used anywhere.

skin/src/main/res/values/strings.xml
dariosalvi78: sign out, not de-registration

skin/src/main/java/org/researchstack/skin/schedule/ScheduleHelper.java
dariosalvi78: this is used by RS to schedule tasks given their scheduling information. I am not sure how it is related to Sage.

skin/src/main/java/org/researchstack/skin/ui/SplashActivity.java
dariosalvi78: I can´t remember now exactly (too long ago!), but basically, without this change the onboarding was buggy. If one stopped the app in the middle of the process the app went into some incoherent state and could not be restored.

backbone/src/main/java/org/researchstack/backbone/task/Task.java
dariosalvi78: PS: I my humble opinion, the whole task / ViewTaskActivity would need a radical rethink...

backbone/src/main/java/org/researchstack/backbone/ui/callbacks/StepViewCallback.java
ghost: I am not sure about the MainThread annotation. The callback is used only in one place: by the ViewTaskActivity, after the animation that goes from one step to another has been completed. It is likely that this happens on the main thread, but there is no special reason why this MUST happen there. Or am I missing something?

skin/src/main/java/org/researchstack/skin/DataProvider.java
ghost: actually adding another method doesn't make much sense, as one would have to implement both methods. I suggest that we keep the method with more parameters in, and let decide to the implementation if all parameters are actually used.

skin/src/main/res/values/strings.xml
ghost: in the case of automatically-generated passwords, as users don't hold the password, sign-out actually means that they will need to re-register.

backbone/src/main/java/org/researchstack/backbone/task/Task.java
liujoshua: Thanks for your input, that makes sense and timed walk is certainly something RS needs to support. Definitely agree a rethink is needed and is something we will be tackling.

skin/src/main/java/org/researchstack/skin/schedule/ScheduleHelper.java
dephillipsmichael: My fault, you don't have access to the Jira obviously.

Just wanted to share we are planning on re-writing the ScheduleHelper soon, so any changes to it should be able to go in.

backbone/src/main/java/org/researchstack/backbone/task/Task.java
dephillipsmichael: I implemented the walking tasks by extending ViewTaskActivity with more functionality.  It seemed to work well without having to store the activity in a model object like Task.

https://github.com/ResearchStack/ResearchStack/pull/349

skin/src/main/java/org/researchstack/skin/DataProvider.java
dephillipsmichael: If you keep the method with the password parameter, please explain the situation thoroughly in the method header and designate the password parameter as @Nullable

backbone/src/main/java/org/researchstack/backbone/task/Task.java
ghost: interesting!
I've found your source step detector algorithm [here](https://github.com/bagilevi/android-pedometer), you'd better test it with a couple of phones at least.

backbone/src/main/java/org/researchstack/backbone/model/DocumentProperties.java
```diff
@@ -88,4 +90,24 @@ public void setRequiresBirthdate(boolean requiresBirthdate)
     {
         this.requiresBirthdate = requiresBirthdate;
     }
+
+    public boolean generatePassword()
+    {
+        return generatePassword;
+    }
+
+    public void setGeneratePassword(boolean generate)
```
dephillipsmichael, line 99: Without any code comments, it is unknown what generate password variable does.  Can you add some comments?

dephillipsmichael:
So, I see this PR has existed for probably way too long lol.  Since it's creation, we have drastically re-architected the entire onboarding process and also laid down the foundation for all Active Tasks.

This also explains the 10+ conflicts that exist now and make this really hard to pull the trigger and just merge in.  My suggestion would be to keep this PR on your own public branch so that your app can target it, but refrain from merging it into RS develop.  Once we get an updated sample app, you will be able to migrate your existing app to RS develop and a lot of the code in this PR will be deprecated.

ghost:
I agree. I see that you are working on a new version of RS, but I am still using the old one, as I started my project even before the first release of RS.

I wish you can improve it, because the version I am using is still very immature. Good luck!

Issue #283, opened 2016-09-27
allow pin to be cleared


dukemike:
In my apps, I would like to clear the pin when the user signs out or withdraws so that I can fully reset the app to its initial state.

dukemike:
PR #286 resolves this issue and was accepted.

bosniltd-a11y:
> In my apps, I would like to clear the pin when the user signs out or withdraws so that I can fully reset the app to its initial state.
>

PR #286, opened 2016-10-05
allow pin to be cleared. refs #283

dukemike:


cseebregts:
Tested functionality on the SampleApp and works great!

mikecarroll:
Thanks @dukemike for PRing and @cseebregts for reviewing!

bosniltd-a11y (approved):

Issue #285, opened 2016-10-04
allow password policy to be set by app developers


dukemike:
Rather than having a fixed password policy in skin's SignUpStepLayout, use the UiManager to allow developers to override the default policy.

PR #287, opened 2016-10-05
allow password policy to be set by app developers. refs #285

dukemike:


liujoshua:
Changing base branch to develop. This is our new preferred flow and avoids needing to bump the version with every merge to master

PR #289, opened 2016-10-12
Fixed result not selected when going back to Duration Question Step

niklasnordmark:
When selecting a duration, going forward to the next step and then pushing back the previously selected values where not show.


jdkizer9:
Thanks @niklasnordmark!

PR #306, opened 2016-12-15
feature/consent_quiz_items_field

dephillipsmichael:
Added "items" parsing to consent json to keep compatibility with iOS.
Moved ConsentQuizModel to backbone, since it is a basic modal class
Also, added unit tests and new question utils class to decouple logic from step layout


EricSiegNW:
Can you move this over to target develop instead?

PR #307, opened 2016-12-15
Feature/gradle updates

EricSiegNW:
Update gradle build tools version
exclude rsb_ prefix so things will build correctly

### DIFF
--- a/backbone/build.gradle
+++ b/backbone/build.gradle
@@ -33,7 +33,7 @@ android {
         exclude 'META-INF/NOTICE.txt'
     }
 
-    resourcePrefix 'rsb_'
+//    resourcePrefix 'rsb_'
 }
 
 Properties properties = new Properties()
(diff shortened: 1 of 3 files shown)


backbone/build.gradle
```diff
@@ -33,7 +33,7 @@ android {
         exclude 'META-INF/NOTICE.txt'
     }
 
-    resourcePrefix 'rsb_'
+//    resourcePrefix 'rsb_'
```
dephillipsmichael: I'm not sure of the root cause of why this is throwing so many warnings.  But, it is actually something we should in the gradle file, maybe @liujoshua knows why its not working correctly?

liujoshua: Are you seeing warnings building just RS? Or in mPower with the submodule structure? If the latter, I have a fix locally I can add here

dephillipsmichael: I believe its when we build mPower

liujoshua: I'm still unable to repro, seems to behave correctly for me. Although I only see these types of errors when running inspections, not during build.

If I rename a resource so it doesn't start with rsb_, I see the warning when I run inspection. Otherwise, looks like all the resources to start with rsb_

Does this issue also occur with rss_ prefix for skin?

PR #308, opened 2016-12-17
Inclusion criteria

rianhouston:
This PR makes changes to read inclusion criteria from JSON rather than hardcoded in UIManager.  It has been tested with the EligibilityRequirements.json from mPower-AppStore and Resilience-ios projects.

### DIFF
--- a/skin/src/main/java/org/researchstack/skin/UiManager.java
+++ b/skin/src/main/java/org/researchstack/skin/UiManager.java
@@ -71,6 +71,7 @@ public abstract class UiManager
      * @param context android context
      * @return a Step used for Eligibility within the onboarding process
      */
+    @Deprecated
     public abstract Step getInclusionCriteriaStep(Context context);
 
     /**
@@ -80,6 +81,7 @@ public abstract class UiManager
      * @param result StepResult object that contains the answers of the InclusionCriteria step
      * @return true if the user is elligible for the study
      */
+    @Deprecated
     public abstract boolean isInclusionCriteriaValid(StepResult result);
 
     /**
--- a/skin/src/main/java/org/researchstack/skin/task/SignUpTask.java
+++ b/skin/src/main/java/org/researchstack/skin/task/SignUpTask.java
@@ -2,11 +2,28 @@ package org.researchstack.skin.task;
 
 import android.content.Context;
 
+import org.researchstack.backbone.answerformat.BooleanAnswerFormat;
+import org.researchstack.backbone.result.StepResult;
 import org.researchstack.backbone.result.TaskResult;
+import org.researchstack.backbone.step.FormStep;
+import org.researchstack.backbone.step.InstructionStep;
+import org.researchstack.backbone.step.QuestionStep;
 import org.researchstack.backbone.step.Step;
+import org.researchstack.backbone.utils.LogExt;
 import org.researchstack.skin.PermissionRequestManager;
+import org.researchstack.skin.R;
+import org.researchstack.skin.ResourceManager;
 import org.researchstack.skin.TaskProvider;
 import org.researchstack.skin.UiManager;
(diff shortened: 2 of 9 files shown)


skin/src/main/java/org/researchstack/skin/UiManager.java
```diff
@@ -71,6 +71,7 @@ public static UiManager getInstance()
      * @param context android context
      * @return a Step used for Eligibility within the onboarding process
      */
+    @Deprecated
```
dephillipsmichael: For documentation purposes, could you  link to the new method in ResourceManager?

skin/src/main/java/org/researchstack/skin/task/SignUpTask.java
```diff
@@ -19,12 +36,82 @@
...
+   */
+  private void initSteps(Context context) {
+        InclusionCriteriaModel model = ResourceManager.getInstance()
+                .getInclusionCriteria()
+                .create(context);
+
+        for(InclusionCriteriaModel.Step s: model.steps) {
+            switch(s.type)
+            {
+                case "instruction":
```
dephillipsmichael: Gson will actually serialize enums automatically if you do @SerializedName above the enum.  This would be a way to avoid hard-coded case values for things like Step.type

skin/src/main/java/org/researchstack/skin/task/SignUpTask.java
```diff
@@ -19,12 +36,82 @@
...
+                    Step instruction = null;
+                    switch(s.identifier)
```
dephillipsmichael: Not sure if "identifier" would make sense for an enum, however.  Can you define a static string somewhere?

skin/src/main/java/org/researchstack/skin/task/SignUpTask.java
```diff
@@ -19,12 +36,82 @@
...
+                // TODO: not sure what the differences are between compound/toggle or is compound obsolete?
+                case "compound":
+                case "toggle":
+                    FormStep form = new FormStep(SignUpInclusionCriteriaStepIdentifier, s.text, s.detailText);
+                    List<QuestionStep> questions = new ArrayList<>();
+
+                    if(s.items != null)
+                    {
+                        // TODO: extend the json to include (yes/no)?
+                        BooleanAnswerFormat booleanAnswerFormat = new BooleanAnswerFormat("Yes", "No");
```
dephillipsmichael: "Yes" and "No" are defined in R.string.rsb_yes and R.string.rsb_no

skin/src/main/java/org/researchstack/skin/task/SignUpTask.java
```diff
@@ -178,4 +265,44 @@ public void setHasPasscode(boolean hasPasscode)
     {
         this.hasPasscode = hasPasscode;
     }
+
+    private Boolean getBooleanAnswer(Map mapStepResult, String id)
+    {
+        StepResult stepResult = (StepResult)mapStepResult.get(id);
```
dephillipsmichael: Nice coverage on this

skin/src/main/res/layout/rss_layout_ineligible.xml
```diff
@@ -13,12 +13,21 @@
...
         android:text="@string/rss_eligibility_you_are_ineligible"
     />
 
+    <TextView
+        android:id="@+id/ineligible_detail"
+        style="@style/Backbone.Survey.Summary"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:textColorLink="?attr/colorAccent"
+        android:textSize="14sp"
```
dephillipsmichael: do we want to define a text style for detail text and use that instead of 14sp?

PR #313, opened 2017-01-03
Feature/ma 11

dephillipsmichael:
This work involves the beginning steps to creating an OnboardingManager that controls the flow of the consent and user onboarding process.  The OnboardingManager operation is customizable through a JSON file provided by the ResourceManager. The JSON specifies the OnboardingSections and their contents. The OnboardingSections themselves are customizable through "steps", or a List of SurveyItems.  The SurveyItem class has an extensive polymorphism structure that allows for many different types of SurveyItems, from BaseSurveyItem to more complex ones like SingleTextChoiceQuestionSurveyItem.

All this work is based off of iOS' swift library BridgeAppSdk's SBAOnboardingManager.  

For reference, the next step in this work, will be creating a fully-functional SurveyFactory that will convert SurveyItems into their designated Step objects.  This paradigm will also be used to parse Surveys from the server.  So, the SurveyFactory and ConsentDocumentFactory are committed but not fully implemented yet.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/model/survey/QuestionSurveyItem.java
+++ b/backbone/src/main/java/org/researchstack/backbone/model/survey/QuestionSurveyItem.java
@@ -1,0 +1,188 @@
+package org.researchstack.backbone.model.survey;
+
+import com.google.gson.annotations.SerializedName;
+
+import org.researchstack.backbone.step.QuestionStep;
+import org.researchstack.backbone.step.Step;
+import org.researchstack.backbone.utils.SurveyFactory;
+
+import java.util.List;
+
+/**
+ * Created by TheMDP on 12/31/16.
+ */
+
+public class QuestionSurveyItem<T extends Object> extends SurveyItem<T> {
+
+    @SerializedName("questionStyle")
+    public boolean questionStyle;
+    @SerializedName("placeholderText")
+    public String placeholderText;
+    @SerializedName("optional")
+    public boolean optional;
+    @SerializedName("range")
+    public SurveyRange range;
+
+    @SerializedName("expectedAnswer")
+    public boolean expectedAnswer; // Does this need to be a generic type?
+
+    public boolean isValidQuestionItem() {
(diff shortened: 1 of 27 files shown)


backbone/src/main/java/org/researchstack/backbone/model/survey/QuestionSurveyItem.java
```diff
@@ -0,0 +1,188 @@
...
+        QuestionStep step = null;
+        if (isBooleanToggle()) {
+            // TODO finish coding
+            //step = new QuestionStep()
+        }
+        return step;
+    }
+
+//
+//    func createFormStep(isSubtaskStep: Bool, factory: SBASurveyFactory? = nil) -> ORKStep {
```
dephillipsmichael: All this will probably be moved into the SurveyFactory, but for now I'm leaving it for reference in where it was in the iOS library.

backbone/src/main/java/org/researchstack/backbone/onboarding/OnboardingSectionAdapter.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public OnboardingSection deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) throws JsonParseException {
+
+        JsonElement typeJson = json.getAsJsonObject().get(OnboardingSection.ONBOARDING_TYPE_GSON);
+        OnboardingSectionType type = context.deserialize(typeJson, OnboardingSectionType.class);
+
+        JsonElement resourceName = json.getAsJsonObject().get(OnboardingSection.ONBOARDING_RESOURCE_NAME_GSON);
+        if (resourceName != null) {
+            // Android does not support spaces or uppercase letters for resource names
```
EricSiegNW: That's something we should maybe try to push back to iOS too and encourage them to stick to naming schemes that work on both platforms w/o adjustment

EricSiegNW:
Great initial work!

PR #317, opened 2017-01-06
Feature/ma 11 factories

dephillipsmichael:
This is a full functional implementation of the Survey Factory and Consent Document Factory.
The unit tests have verified that it can parse the iOS version of the Onboarding.json, which also parses consent.json and eligibility.json and create a NavigableOrderedTask full of the correct steps that will guide the user correctly through on boarding by using JSON.

The SurveyFactory will also be what is used for parsing surveys that come down from the loadTask web call once we get web survey functionality into the app.

I understand this is a lot of code, but know that most of it was transcribed from the iOS BridgeAppSdk


backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItem.java
```diff
@@ -54,4 +51,30 @@ public int compare(SurveyItemType lhs, SurveyItemType rhs) {
             return 0;
         }
     }
+
+    @Override
+    public int hashCode() {
+        if (identifier == null) {
```
liujoshua: Why is identifier the only field we need?

backbone/src/main/java/org/researchstack/backbone/answerformat/GenderAnswerFormat.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public GenderAnswerFormat(Context context) {
+        super(ChoiceAnswerStyle.SingleChoice, createChoices(context));
+    }
+
+    /**
+     * @param context used to build strings for male, female, other
+     * @return Choice array for gender
+     */
+    static Choice[] createChoices(Context context) {
```
liujoshua: Should we make Choice type parameterized? (I see ChoiceAnswerFormat is where that gets removed)

backbone/src/main/java/org/researchstack/backbone/model/ConsentDocument.java
```diff
@@ -66,6 +68,29 @@
...
+     */
+    @SerializedName("requiresBirthdate")
+    private boolean requiresBirthdate = true;
+    public boolean getRequiresBirthdate() { return requiresBirthdate; }
+    public void setRequiresBirthdate(boolean requiresBirthdate) { this.requiresBirthdate = requiresName; }
+
+    /**
+     * True if consent must require the user's signature
+     */
+    @SerializedName("requiresSignature")
```
liujoshua: Nit-picky comment: let's make the organization consistent. Either fields all at top, getter/setters below. Or grouped like you have it.

backbone/src/main/java/org/researchstack/backbone/model/ProfileInfoOption.java
```diff
@@ -0,0 +1,101 @@
...
+            ProfileInfoOption option = toProfileInfoOption(identifier);
+            if (option != null) {
+                options.add(option);
+            }
+        }
+        return options;
+    }
+
+    public static ProfileInfoOption toProfileInfoOption(String identifier) {
+        List<ProfileInfoOption> options = new ArrayList<>();
```
dephillipsmichael: I'm going to switch this to use Gson

dephillipsmichael: done

backbone/src/main/java/org/researchstack/backbone/model/ConsentDocument.java
dephillipsmichael: I agree, fixing...

dephillipsmichael: fixed

backbone/src/main/java/org/researchstack/backbone/model/ProfileInfoOption.java
```diff
@@ -0,0 +1,101 @@
...
+    @SerializedName("wakeTime")
+    WAKE_TIME("wakeTime"),
+    @SerializedName("sleepTime")
+    SLEEP_TIME("sleepTime");
+
+    String identifier;
+    public String getIdentifier() { return identifier; }
+
+    /** Only useful for when Enum == PASSWORD */
+    boolean addConfirmPassword = false;
```
dephillipsmichael: I remember having trouble keeping it in SurveyFactory, but let me take a second look.

dephillipsmichael: Good find on this, definitely shouldn't be in the enum.  I was able to move it to be a parameter within SurveyFactory.

backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItem.java
dephillipsmichael: Good question... I wish I had a better answer than, "its all I needed to get unit tests and the code working"

What else do you think would be appropriate for comparison?

backbone/src/main/java/org/researchstack/backbone/model/ProfileInfoOption.java
```diff
@@ -0,0 +1,101 @@
...
+    public void setAddConfirmPassword(boolean addConfirmPassword) {
+        this.addConfirmPassword = addConfirmPassword;
+    }
+    public boolean getAddConfirmPassword() {
+        return addConfirmPassword;
+    }
+
+    ProfileInfoOption(String identifier) {
```
dephillipsmichael: screen shot 2017-01-09 at 6 37 41 pm

backbone/src/main/java/org/researchstack/backbone/answerformat/GenderAnswerFormat.java
dephillipsmichael: I'm not sure what you mean.  Do you mean like a new constructor in ChoiceAnswerFormat?  
For more context, I made this method a static method to avoid getting each String twice from Context when making the Choice objects.

backbone/src/main/java/org/researchstack/backbone/model/survey/QuestionSurveyItem.java
```diff
@@ -38,69 +32,13 @@ public boolean isCompoundStep() {
...
-//    }
-
-    public Step createQuestionStep(boolean isSubtaskStep, SurveyFactory factory) {
-        // Factory method for determining the proper type of form-style step to return
-        // the ORKQuestionStep and ORKFormStep have a different UI presentation
-        QuestionStep step = null;
-        if (isBooleanToggle()) {
-            // TODO finish coding
-            //step = new QuestionStep()
+    public boolean usesNavigation() {
```
dephillipsmichael: added

backbone/src/main/java/org/researchstack/backbone/model/ProfileInfoOption.java
```diff
@@ -0,0 +1,101 @@
...
+    @SerializedName("height")
+    HEIGHT("height"),
+    @SerializedName("weight")
+    WEIGHT("weight"),
+    @SerializedName("wakeTime")
+    WAKE_TIME("wakeTime"),
+    @SerializedName("sleepTime")
+    SLEEP_TIME("sleepTime");
+
+    String identifier;
```
dephillipsmichael: good call, done.

backbone/src/main/java/org/researchstack/backbone/model/survey/NavigationStep.java
```diff
@@ -0,0 +1,24 @@
...
+import org.researchstack.backbone.result.TaskResult;
+import org.researchstack.backbone.step.QuestionStep;
+
+import java.util.List;
+
+/**
+ * Created by TheMDP on 12/31/16.
+ */
+
+public interface NavigationStep {
```
dephillipsmichael: This is an unusual interface to have getters/setter like this.  I made it to keep the code similar to the iOS Swift version.  In that language you can member variables in interfaces.

backbone/src/main/java/org/researchstack/backbone/result/TaskResult.java
```diff
@@ -27,10 +34,24 @@
     // unimplemented but exists in RK, implement or delete if not needed
     private Uri outputDirectory;
 
+    /* Default identifier for serilization/deserialization */
+    TaskResult() {
+        super();
+        this.results = new LinkedHashMap<>();
```
dephillipsmichael: LinkedHashMaps retain order of insertion, which the iOS algorithms use sometimes.  So I figured it would be best to keep the results ordered if we ever iterate them.

backbone/src/main/java/org/researchstack/backbone/model/survey/NavigationStep.java
liujoshua: Would it help to make this an abstract class? Or does this interface need to support multiple inheritance?

backbone/src/main/java/org/researchstack/backbone/answerformat/GenderAnswerFormat.java
liujoshua: I was wondering if we should make the return type Choice<T>[]

backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItem.java
liujoshua: One one hand, the models are probably supposed to be uniquely identifiable by the identifier, but on the other hand, client/server might get out of sync, and having equals take into account the fields could make this syncing easier.

For unit tests, I find it useful to include all the persisted fields (ones that are serialized/deserialized) for models. Calling deepCopy() should product an object that is equal. 

Whether we're comparing all persisted fields or comparing the identifier (which I assume should be globally unique) we should make that consistent across models.

backbone/src/main/java/org/researchstack/backbone/answerformat/GenderAnswerFormat.java
dephillipsmichael: instead of static Choice[]?  unfortunately you can't call an object's methods before its super constructor has been called, so it needs to be static.

backbone/src/main/java/org/researchstack/backbone/model/survey/NavigationStep.java
dephillipsmichael: Yes, the steps need multiple inheritance, since they usually extend some other step, and then also need to add NavigationStep support.

backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItem.java
dephillipsmichael: Those are all good thoughts.  

I also think that "identifier" was meant to the on unique item, and I believe that is enforced on the bridge study manager.

After reflecting more, I believe the best approach here would be to keep the current implementation, and then address adding more fields when we implement the load survey task.  Because, once that is implemented, we should have a better feel for the usefulness of loading SurveyItems, saving them in the DB, and who knows, maybe we don't end up even saving the SurveyItems but saving the produced Task/Step relationships instead.  Does that sound like a good approach?  If so, I can add some TODO: comments here and maybe you can add some in the Bridge API to address this.

backbone/src/main/java/org/researchstack/backbone/answerformat/GenderAnswerFormat.java
liujoshua: Oops, I meant to type Choice<T>[]

backbone/src/main/java/org/researchstack/backbone/model/GsonSerializablePolymorphism.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /*
+     * THE ORDER OF THE DATA PAIRS IS VERY IMPORTANT
+     *
+     * The deepest nested GsonSerializablePolymorphism class pair must go first,
+     * With your base class last
+     *
+     * The base/sub class pairs to use in the automatic polymorphism serialization/deserialization
+     * For example, if you have a base class Shape.class, and a Triangle.class that extends Shape.class,
+     * You would do...
```
liujoshua: Which class are you calling getClass() on?

backbone/src/main/java/org/researchstack/backbone/model/GsonSerializablePolymorphism.java
```diff
@@ -0,0 +1,106 @@
...
+     * List A = new List { new GsonBaseSubClassPair(Shape.class, getClass()) }
+     * return new Data(Shape.class, A);
+     *
+     * If Shape.class has member which LineType.class base, and its of subclass type BigLineType.class
+     * You would return something like this,
+     * List A = new List { new GsonSerializablePolymorphism.DataPair(LineType.class, BigLineTypeObj.getClass()) }
+     * A.add(new GsonSerializablePolymorphism.DataPair(LineType.class, BigLineTypeObj.getClass()));
```
liujoshua: Is A really supposed to contain the same DataPair twice?

dephillipsmichael: No, it's not.  Good find.  Fixed it, take another look.

backbone/src/main/java/org/researchstack/backbone/model/GsonSerializablePolymorphism.java
dephillipsmichael: No, it's not.  Good find.  Fixed it, take another look.

backbone/src/main/java/org/researchstack/backbone/utils/ObjectUtils.java
```diff
@@ -0,0 +1,39 @@
...
+    * Performs a deep copy on the object of type using Gson
+    * Object type must have a default constructor to work properly
+    *
+    * NOTE: this does not work with Polymorphism
+    * Use GsonSerializablePolymorphism class instead
+    */
+    public static <T> T deepCopy(Object object, Class<T> type, Gson gson) {
+        try {
+            String copyJson = gson.toJson(object, type);
+            return gson.fromJson(copyJson, type);
```
liujoshua: In deepCopy and `DataPair` and` Data`, do we want to use java.lang.reflect.Type instead of `Class<T>`?

I think by using `Class<T>` we lose generic type information, so we should use this method instead
https://google.github.io/gson/apidocs/com/google/gson/Gson.html#fromJson-java.lang.String-java.lang.reflect.Type-

backbone/src/main/java/org/researchstack/backbone/model/GsonSerializablePolymorphism.java
```diff
@@ -0,0 +1,106 @@
...
+        // Now make the type adpater for the base most class
+        return ObjectUtils.deepCopy(this, data.baseClass, gson);
+    }
+
+    TypeAdapter createTypeAdapter(Gson gson, DataPair pair) {
+        if (pair.baseClass == pair.subClass) {
+            return null;   // no type adapter needed
+        }
+        // "class" just cant be the name of a SerializedName in this class, which it is unlikely it would
+        RuntimeTypeAdapterFactory typeFactory = RuntimeTypeAdapterFactory.of(pair.baseClass, "class");
```
liujoshua: TypeAdapterFactory can produce a different adapters, so seems like each Data could get one RuntimeTypeAdapterFactory, and each DataPair should get a subtype.

RuntimeTypeAdapterFactory lineFactory = RuntimeTypeAdapterFactory.of(LineType.class).registerSubtype(BigLineTypeObj.class);
RuntimeTypeAdapterFactory shapeFactory = RuntimeTypeAdapterFactory.of(Shape.class).registerSubtype(TriangleObj.class);

Gson gson = new GsonBuilder().registerTypeAdapter(shapeFactory).registerTypeAdapter(lineFactory).build();

I would expect this gson instance to be able to serialize the subclasses correctly (including Shape's member Line)

We should be able to build Gson once, and use the same instance for doing deepCopy for all objects it knows about.

backbone/src/main/java/org/researchstack/backbone/answerformat/TextAnswerFormat.java
```diff
@@ -1,16 +1,21 @@
...
 {
     public static final int UNLIMITED_LENGTH = 0;
-    private int maximumLength;
+    int maximumLength;
+    int minimumLength = 0;
 
-    private boolean isMultipleLines = false;
+    boolean isMultipleLines = false;
+    int     inputType       = InputType.TYPE_CLASS_TEXT;
+    String  validationRegex = null;
```
liujoshua: should we keep a Pattern or Matcher instead of a string?

backbone/src/main/java/org/researchstack/backbone/answerformat/TextAnswerFormat.java
```diff
@@ -1,16 +1,21 @@
...
+
 /**
  * The TextAnswerFormat class represents the answer format for questions that collect a text
  * response from the user.
  */
 public class TextAnswerFormat extends AnswerFormat
 {
     public static final int UNLIMITED_LENGTH = 0;
-    private int maximumLength;
+    int maximumLength;
```
liujoshua: Why not keep the fields private?

backbone/src/main/java/org/researchstack/backbone/model/survey/NavigationStep.java
liujoshua: Do the classes using NavigationStep need to set these values, or just retrieve them? I'm thinking maybe some NavigationSteps have skip logic that is more complicated, do you know if that's the case?

backbone/src/main/java/org/researchstack/backbone/model/GsonSerializablePolymorphism.java
dephillipsmichael: Ok... I ended up scrapping all of GsonSerializablePolymorphism.  I realize now it was an over ambition approach to solving a problem that already has a good solution built into the language.  

I ended up implementing Serializable for all the objects that will be cloned, and whatever member variables they have, must implement Serializable as well.  Also, now every subclass or class that implements Serializable, must have a default constructor, as per the rules of the interface.

I verified all the unit tests still work, which means the object cloning is also working.  Specifically the unit test, NavigableOrderedTaskTest.

backbone/src/main/java/org/researchstack/backbone/answerformat/TextAnswerFormat.java
dephillipsmichael: My default is to keep the language default protected, I guess it's just habit.  But I understand the benefit in a library to keep member variables private as well.  I'll switch these to private.

backbone/src/main/java/org/researchstack/backbone/answerformat/TextAnswerFormat.java
dephillipsmichael: That's a good question. My vote is to stick with validationRegex to keep parity with iOS.

backbone/src/main/java/org/researchstack/backbone/model/survey/NavigationStep.java
dephillipsmichael: The classes using NavigationStep I think just retrieve them.  And there is definitely more complex skip logic used by the NavigableOrderedTask in the iOS library; however, this is sufficient for now. 

 I'm thinking this NavigationStep interface will need iterated on when I start fully implementing the Steps in the on boarding process, and the UI they relate to.  I'll put a TODO in this class that it needs expanded to support SBANavigationRule, SBAConditionalRule, and SBANavigationSkipRule in the near-future.

backbone/src/main/java/org/researchstack/backbone/step/QuestionStep.java
```diff
@@ -26,6 +31,11 @@
 
     private String placeholder;
 
+    /* Default constructor needed for serilization/deserialization of object */
```
liujoshua: (Comment is about this class, but not code you changed)

Why does getStepLayoutClass() return a `Class`, and getStepBodyClass() return a `Class<?>`

backbone/src/main/java/org/researchstack/backbone/model/ConsentQuizModel.java
```diff
@@ -18,6 +18,10 @@
     private String incorrectIcon = "rsb_quiz_retry";
     private String correctIcon   = "rss_ic_quiz_valid";
 
+    ConsentQuizModel() {
```
dephillipsmichael: Yes and also because Model objects should at least have package level default constructors.

backbone/src/main/java/org/researchstack/backbone/model/ConsentSection.java
```diff
@@ -132,6 +146,9 @@ public Type getType()
     {
         return type;
     }
+    void setType(Type type) {
```
dephillipsmichael: No, type should only be set in ConsentSectionAdapter

backbone/src/main/java/org/researchstack/backbone/model/ConsentSection.java
```diff
@@ -148,6 +165,10 @@ public String getCustomImageName()
         return customImageName;
     }
 
+    void setCustomImageName(String imageName) {
```
dephillipsmichael: same answer as about setType

backbone/src/main/java/org/researchstack/backbone/model/ConsentSection.java
```diff
@@ -185,6 +206,19 @@ public String getCustomLearnMoreButtonTitle()
...
+
+    public String getTypeIdentifier() {
+        if (type == Type.Custom) {
+            return customTypeIdentifier;
+        }
+        return type.getIdentifier();
+    }
+
+    public static final int UNDEFINED_RES = -1;
     public enum Type implements Serializable
```
dephillipsmichael: Good point.  Yeah, all of ConsentDocument was written already in master.

backbone/src/main/java/org/researchstack/backbone/model/ConsentSectionAdapter.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Created by TheMDP on 1/5/17.
+ */
+
+public class ConsentSectionAdapter implements JsonDeserializer<ConsentSection> {
+
+    /**
+     * Used to convert ConsentSections
+     */
+    Context androidContext;
```
dephillipsmichael: Sure? :P doesn't matter really

backbone/src/main/java/org/researchstack/backbone/model/ProfileInfoOption.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.List;
+
+/**
+ * Created by TheMDP on 1/4/17.
+ *
+ * Used by several differnt Step types to designate re-usable QuestionStep types
+ * that collect user profile info
+ */
+
+public enum  ProfileInfoOption {
```
dephillipsmichael: I'd rather stick with keeping parity with iOS.  But i'm open to discussion.

backbone/src/main/java/org/researchstack/backbone/onboarding/CustomOnboardingSection.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class CustomOnboardingSection extends OnboardingSection {
+
+    CustomOnboardingSection(String customOnboardingType) {
+        this.customOnboardingType = customOnboardingType;
+    }
+
+    @Override
+    public String getOnboardingSectionIdentifier() {
+        if (onboardingType == OnboardingSectionType.CUSTUM) {
```
dephillipsmichael: whoops... fixing

backbone/src/main/java/org/researchstack/backbone/onboarding/OnboardingSection.java
```diff
@@ -14,11 +15,34 @@
 
 public class OnboardingSection {
 
-    public OnboardingSection() {}
+    /**
+     * These are pre-defined onboarding section identifiers
+     * that are used to create default helper enums for OnboardingSectionType
+     */
+    public static final String LOGIN_IDENTIFIER = "login";
```
dephillipsmichael: In the past in iOS, Eric and I have needed to access these identifiers, so I think it' valid to make them public.  

Also, I had to make OnboardingSectionType.LOGIN.getIdentifier() a protected method, because you shouldn't call it.  You should call getTypeIdentifier() on the OnboardinSection object instead. I changed it to this to support unique custom identifiers per enum.

backbone/src/main/java/org/researchstack/backbone/step/QuestionStep.java
dephillipsmichael: Not sure, I didn't write this class.  Do you think it needs changed?

backbone/src/main/java/org/researchstack/backbone/step/ToggleFormStep.java
```diff
@@ -0,0 +1,23 @@
+package org.researchstack.backbone.step;
+
+import java.util.List;
+
+/**
+ * Created by TheMDP on 1/3/17.
+ */
+
+public class ToggleFormStep extends NavigationFormStep {
```
dephillipsmichael: I think this just needs better documentation

dephillipsmichael: because its different in the form of what it's question steps are, only boolean questions, which is controlled by survey factory injecting boolean steps into the constructor

backbone/src/main/java/org/researchstack/backbone/answerformat/GenderAnswerFormat.java
dephillipsmichael: static methods don't support generic typing

backbone/src/main/java/org/researchstack/backbone/answerformat/TextAnswerFormat.java
```diff
@@ -40,6 +46,32 @@ public int getMaximumLength()
         return maximumLength;
     }
 
+    /**
+     * Set the minimum length for the answer, 0 if no minumum set
```
EricSiegNW: c/p typo, should be maximum length

backbone/src/main/java/org/researchstack/backbone/answerformat/PasswordAnswerFormat.java
```diff
@@ -0,0 +1,28 @@
...
+
+import android.text.InputType;
+
+/**
+ * Created by TheMDP on 1/4/17.
+ */
+
+public class PasswordAnswerFormat extends TextAnswerFormat {
+
+    static final int DEFAULT_PASSWORD_MIN_LENGTH    = 4;
```
EricSiegNW: don't think they should be private, would want a single source of "min" and "max" that can be referenced

backbone/src/main/java/org/researchstack/backbone/model/ConsentSection.java
```diff
@@ -278,81 +344,37 @@ public String getCustomLearnMoreButtonTitle()
          * property).
          */
         @SerializedName("onlyInDocument")
-        OnlyInDocument;
+        OnlyInDocument("onlyInDocument", UNDEFINED_RES, R.string.rsb_consent_section_more_info, null);
 
-        public int getTitleResId()
+        Type(
+            String identifier,
```
EricSiegNW: nice cleanup

backbone/src/main/java/org/researchstack/backbone/onboarding/ConsentOnboardingSection.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Created by TheMDP on 1/5/17.
+ */
+
+public class ConsentOnboardingSection extends OnboardingSection {
+
+    // Serialization must be done manually in OnboardingSectionAdapter
+    ConsentDocument consentDocument;
+
+    public SurveyFactory getDefaultOnboardingSurveyFactory(Context context) {
```
EricSiegNW: ^

backbone/src/main/java/org/researchstack/backbone/step/InstructionStep.java
```diff
@@ -1,16 +1,76 @@
...
+
 /**
  * An InstructionStep object gives the participant instructions for a task.
  * <p>
  * You can use instruction steps to present various types of content during a task, such as
  * introductory content, instructions in the middle of a task, or a final message at the completion
  * of a task.
  */
-public class InstructionStep extends Step
+public class InstructionStep extends Step implements NavigableOrderedTask.NavigationRule
```
EricSiegNW: General FYI question: what would be the process to subclass InstructionStep w/ extra fields and possibly a different UI and then ensure that is used instead of InstructionStep?

backbone/src/main/java/org/researchstack/backbone/task/NavigableOrderedTask.java
```diff
@@ -0,0 +1,404 @@
...
+import java.util.Map;
+
+/**
+ * Created by TheMDP on 12/29/16.
+ *
+ * TODO this class needs expanded to support
+ * TODO SBANavigationRule, SBAConditionalRule, and SBANavigationSkipRule in the near-future.
+ */
+
+public class NavigableOrderedTask extends OrderedTask implements TaskResultSource {
```
EricSiegNW: Really cool class :)

backbone/src/test/java/org/researchstack/backbone/answerformat/AnswerFormatTest.java
```diff
@@ -39,4 +39,33 @@ public void testTestInvalidEmailAnswerFormat() throws Exception
...
+        assertTrue(format.isAnswerValid("Abcd"));               // 4 characters is valid up to
+    }
+
+    @Test
+    public void testPasswordAnswerFormat()
+    {
+        PasswordAnswerFormat format = new PasswordAnswerFormat();
+        assertTrue(format.isAnswerValid("Abcd1234"));           // normal password valid
+        assertTrue(format.isAnswerValid("Abcd"));               // 4 characters is valid up to
+        assertTrue(format.isAnswerValid("Abcd1234Abcd1234"));   // 16 characters
```
EricSiegNW: other requirements that bridge can require: includes 1 number, includes 1 lower case, 1 upper case, 1 symbol. These are optional tho.

Where did the max password length come from? I don't think that's a bridge thing.

dephillipsmichael: Good point, these rules are sort of Bridge coupled, do you think I should just switch it to any length?  For max length see makeConfirmationFormItem method in SBAProfileInfoForm.

backbone/src/main/java/org/researchstack/backbone/step/InstructionStep.java
dephillipsmichael: Good question. I really make it obvious in the next PR, stay tuned.

backbone/src/main/java/org/researchstack/backbone/model/ConsentSection.java
dephillipsmichael: thanks :)

backbone/src/main/java/org/researchstack/backbone/answerformat/TextAnswerFormat.java
dephillipsmichael: fixed in other PR

backbone/src/test/java/org/researchstack/backbone/answerformat/AnswerFormatTest.java
EricSiegNW: yeah, they are doing 4/16 as you did so we can just stay synced with them. 4 is the default min in bridge anyway and I've never seen it go lower, not sure what logic resulted in the 16 char max but seems easy to just implement it here as well.

PR #318, opened 2017-01-06
Changed initialization of AppPrefs to not be so lazy

rianhouston:


skin/src/main/res/values/strings.xml
```diff
@@ -113,7 +113,7 @@
     <string name="rss_settings_general_software_notices">Software Notices</string>
     <string name="rss_settings_general_leave_study">Leave Study</string>
     <string name="rss_settings_general_join_study">Join Study</string>
-    <string name="rss_settings_version">Build number %1$s (%2$s)</string>
+    <string name="rss_settings_version">Build number %1$s (%2$d)</string>
```
rianhouston: For some reason this just started showing as an error for me even though it hasn't changed.  Version code is an integer and not a string though so this is correct.

dephillipsmichael: Did you update studio maybe?

dephillipsmichael:
Looks good! approved.  Eric will have to merge

skin/src/main/java/org/researchstack/skin/AppPrefs.java
```diff
@@ -24,11 +24,16 @@
         prefs = PreferenceManager.getDefaultSharedPreferences(context);
     }
 
-    public static synchronized AppPrefs getInstance(Context context)
+    public static void init(Context context) {
```
liujoshua: getInstance() is synchronized, and it doesn't seem the instance variable is used in a thread safe manner in this class

skin/src/main/java/org/researchstack/skin/AppPrefs.java
```diff
@@ -24,11 +24,16 @@
...
+        instance = new AppPrefs(context);
+    }
+
+    public static synchronized AppPrefs getInstance()
```
liujoshua: Do we want to keep this class backwards compatible and mark the old one as `@Deprecated`?

I do like this change -- seems like a bad idea to require context to be passed every time since most of the time the context is ignored. 

Or, if the goal is to simply force loading of AppPrefs at a certain point, we could call AppPrefs.getInstance(context) in ResearchStack.init

skin/src/main/java/org/researchstack/skin/AppPrefs.java
liujoshua: Could you please look at this comment prior to merge?

skin/src/main/java/org/researchstack/skin/AppPrefs.java
dephillipsmichael: Yeah, I think we should add call to AppPrefs.getInstance(context) in ResearchStack.init
and I'm always for deprecating over removing in an SDK.

skin/src/main/java/org/researchstack/skin/AppPrefs.java
dephillipsmichael: As long as init is called from the main thread, i don't think we would hit thread issues.  SharedPrefs is thread-safe

liujoshua: That's a very good point. Should we remove `synchronized` from getInstance since it's not needed, and to make it clearer it's not something we need to worry about in future changes to this class?

rianhouston:
Thanks for the feedback guys.  I added back the original method and marked as deprecated.  Also removed synchronized from the new method.

skin/src/main/java/org/researchstack/skin/AppPrefs.java
```diff
@@ -24,6 +23,11 @@
         prefs = PreferenceManager.getDefaultSharedPreferences(context);
     }
 
+    public static void init(Context context) {
+        instance = new AppPrefs(context);
+    }
+
+    @Deprecated
     public static synchronized AppPrefs getInstance(Context context)
```
liujoshua: Nit-pick: We could remove synchronized to make it cleare that synchronize isn't needed or properly implemented and it won't affect the method signature for backwards compatibility purposes.

PR #319, opened 2017-01-19
Feature/onboarding ui

dephillipsmichael:


backbone/src/main/javaorg/researchstack/backbone/DataProvider.java
```diff
@@ -160,6 +188,15 @@ public static void init(DataProvider instance)
     public abstract void saveConsent(Context context, TaskResult consentResult);
```
dephillipsmichael: My initial thought was to deprecate this, since it encourages coupling the DataProvider on the StepLayouts. What do you guys think?

backbone/src/main/javaorg/researchstack/backbone/DataProvider.java
```diff
@@ -169,6 +206,13 @@ public static void init(DataProvider instance)
     public abstract User getUser(Context context);
 
     /**
+     * Saves the user object
+     * @param context android context
+     * @param user User object to save
+     */
+    public abstract void setUser(Context context, User user);
```
dephillipsmichael: I thought it was odd that this wasn't an abstract method, but I I think it was because all User object was before was name, email, birthday, which were set automatically by other methods in DataProvider.

backbone/src/main/javaorg/researchstack/backbone/model/User.java
```diff
@@ -0,0 +1,85 @@
+package org.researchstack.backbone.model;
+
+import java.io.Serializable;
+import java.util.Date;
+
+/*
+  Created by bradleymcdermott on 10/22/15.
+ */
+public class User implements Serializable
```
dephillipsmichael: I'm not sure why the git history didn't transfer for this file, but not sure it is that important.

backbone/src/main/java/org/researchstack/backbone/PermissionRequestManager.java
```diff
@@ -0,0 +1,293 @@
...
+ * This class is responsible for checking permissions during runtime. It is flexible enough to
+ * handle system and custom permissions (like gainging auth for a 3rd party service).
+ * <p>
+ * Any PermissionRequest should be added to the PermissionRequestManager during {@link
+ * Application#onCreate()}
+ * <p>
+ * The PermissionRequest objects are then presented in {@link PermissionStepLayout} for denile /
+ * approval
+ */
+public abstract class PermissionRequestManager
```
dephillipsmichael: Hmmm, I don't like how the history left this file in particular.  I've always had trouble when moving a file around and changing its contents in the same PR.  I'll try and fix this.

backbone/src/main/java/org/researchstack/backbone/answerformat/EmailAnswerFormat.java
```diff
@@ -1,15 +1,19 @@
...
 
 public class EmailAnswerFormat extends TextAnswerFormat
 {
     private static final int MAX_EMAIL_LENGTH = 255;
 
+    /* Default constructor needed for serilization/deserialization of object */
     public EmailAnswerFormat()
     {
         super(MAX_EMAIL_LENGTH);
+        setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS);
```
dephillipsmichael: Thought it was odd, but you can't just have InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS without also having InputType.TYPE_CLASS_TEXT.   Same goes for any of the variations of TYPE_NUMBER

dephillipsmichael:
This is too confusing to review without the other feature PR being merged in first, closing for now.

PR #320, opened 2017-01-19
Feauture/onboarding ui2

dephillipsmichael:
This is the final PR for the new OnboardingManager implementation.

It implements the StepLayouts for the corresponding Steps built in both SurveyFactorys. Most of the StepLayouts were reused from Skin, but of note, FormStepLayout, ProfileStepLayout, and SublistStepLayout are new generic StepLayouts that we can use.

This also adds the hooks for using the OnboardingManager.

This also fixes a few bugs in NavigableOrderedTask as it relates to the Navigation paradigm.  These bugs were more obvious as I began to test the Onboarding UI flow.


backbone/src/main/java/org/researchstack/backbone/DataProvider.java
```diff
@@ -143,13 +160,24 @@ public static void init(DataProvider instance)
...
     /**
      * This method is responsible in uploading the user consent information (e.g. Name, Birthdate,
-     * Signature) to the backend
+     * Signature) to the backend.  Usually, this is done by looking into the TaskResult
+     * object and filling up the ConsentSignature and then calling the method below this
+     * with the signature parameter
      *
      * @param context android context
      */
     public abstract void uploadConsent(Context context, TaskResult consentResult);
```
dephillipsmichael: My initial thought was to deprecate this method, since it encourages coupling the DataProvider with the Results of the StepLayouts. What do you guys think?

backbone/src/main/java/org/researchstack/backbone/answerformat/EmailAnswerFormat.java
```diff
@@ -11,6 +13,7 @@
     public EmailAnswerFormat()
     {
         super(MAX_EMAIL_LENGTH);
+        setInputType(InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS);
```
dephillipsmichael: Thought it was odd, and wanted to call it out for you guys, but you can't just have InputType.TYPE_TEXT_VARIATION_EMAIL_ADDRESS without also having InputType.TYPE_CLASS_TEXT. Same goes for any of the variations of TYPE_NUMBER

backbone/src/main/java/org/researchstack/backbone/model/ConsentSignatureBody.java
```diff
@@ -0,0 +1,45 @@
+package org.researchstack.backbone.model;
```
dephillipsmichael: I think the git history was cleared from this file, but I don't imagine it is important.  Also, not sure i'll be able to revive it.

backbone/src/main/java/org/researchstack/backbone/result/TaskResultSource.java
```diff
@@ -1,19 +0,0 @@
-package org.researchstack.backbone.result;
-
-/**
- * Created by TheMDP on 12/30/16.
- */
-
-public interface TaskResultSource {
```
dephillipsmichael: this was something i coded from iOS, that didn't end up being used

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/PasscodeCreationStepLayout.java
```diff
@@ -0,0 +1,208 @@
...
+import org.researchstack.backbone.StorageAccess;
+import org.researchstack.backbone.result.StepResult;
+import org.researchstack.backbone.result.TaskResult;
+import org.researchstack.backbone.step.PasscodeStep;
+import org.researchstack.backbone.step.Step;
+import org.researchstack.backbone.ui.callbacks.StepCallbacks;
+import org.researchstack.backbone.ui.views.PinCodeLayout;
+import org.researchstack.backbone.utils.ThemeUtils;
+
+public class PasscodeCreationStepLayout extends PinCodeLayout implements StepLayout
```
dephillipsmichael: This is essentially the same class of SignUpPincodeCreationStepLayout; however, I want to deprecate that moving forward, as this will become the passcode creation layout.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/StepLayout.java
```diff
@@ -1,13 +1,23 @@
...
 {
-    void initialize(Step step, StepResult result);
+    /**
+     * @param step Step to be related to this StepLayout
+     * @param result the StepResult for this step, if one already exists
+     * @param taskResult The TaskResult object, if this StepLayout belongs to a Task
+     */
+    void initialize(@NonNull Step step,
+                    @Nullable StepResult result,
+                    @Nullable TaskResult taskResult);
```
dephillipsmichael: Definitely want some opinions on this change.  I thought it would be helpful that the TaskResult be passed around so the Steps know about the results of previous Steps.  This is one way we can keep from saving the user's password, to use it to verify email, without exposing a security flaw.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ConsentReviewSubstepListStepLayout.java
```diff
@@ -0,0 +1,128 @@
...
+            DataProvider.getInstance().saveConsent(getContext(), consentSignatureBody);
+            super.onComplete();
+        }
+    }
+
+    protected void uploadConsent(ConsentSignatureBody consentSignatureBody) {
+        DataProvider.getInstance()
+                .uploadConsent(getContext(), consentSignatureBody)
+                .compose(ObservableUtils.applyDefault())
+                .subscribe(dataResponse -> {
```
dephillipsmichael: I think this may crash if the activity is destroyed while it is running.  I'm going to work on a generic way to check if context is still valid.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/StepLayout.java
EricSiegNW: I think this is cool, something for people to be aware of tho so we can avoid concurrency issues or unexpected changes

backbone/src/main/java/org/researchstack/backbone/answerformat/TextAnswerFormat.java
```diff
@@ -47,7 +47,7 @@ public int getMaximumLength()
     }
 
     /**
-     * Set the minimum length for the answer, 0 if no minumum set
+     * Set the maximum length for the answer, 0 if no maximum set
```
dephillipsmichael: Yeah, check the isValid method

backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItem.java
```diff
@@ -35,18 +35,6 @@
...
     public List<T> items;
 
-    @SerializedName("skipIdentifier")
-    public String skipIdentifier;
-
-    @SerializedName("skipIfPassed")
-    public boolean skipIfPassed;
-
-    // TODO: implement this?
-    public String rulePredicate;  // this is an NSPredicate on iOS, how do we convert?
```
dephillipsmichael: I realized there is no such thing as a predicate on Android, so I removed that field.  But the logic behind what it means to have a predicate was moved into expectedAnswer, which you can see in NavigationQuestionStep.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ConsentReviewSubstepListStepLayout.java
liujoshua: Maybe you could keep track of the subscription and call unsubscribe in onStop

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/StepLayout.java
liujoshua: From what I can gather, Step comes from Task (getStepAfterStep, etc), so I'm wondering if the step being passed in could already have the relevant data set on it. Passing taskResult seems to ask a step to understand what a Task is, and what other Steps are, and so far it looks like Task knows about Step, not as much the other way around.

Another thing I'm thinking is that StepLayout has a callback, but a PR also recently added a callback on Task. Task could then be responsible for moving data around between steps.

I'll definitely defer to you guys since you have better context on what the use cases are and what RK does

backbone/src/main/java/org/researchstack/backbone/utils/StepLayoutHelper.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.reflect.Constructor;
+
+/**
+ * Created by TheMDP on 1/16/17.
+ */
+
+public class StepLayoutHelper {
+
+    @NonNull
+    @MainThread
```
liujoshua: As it is now, does this method need to be called on MainThread?

dephillipsmichael: Yes, you're not allowed to instantiate any Views on a thread besides Main.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/StepLayout.java
dephillipsmichael: Those are some good thoughts Josh.  I think I'm going to revert this change, and push some more logic into the OnboardingTaskActivity.  That activity seems like the one that should be passing around certain info to it's StepLayouts, and it is also the one that knows the TaskResult.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ProfileStepLayout.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+        if (result == null) {
+            result = new StepResult<StepResult>(step);
+        }
+        for (ProfileInfoOption option : getProfileStep().getProfileInfoOptions()) {
+            // Look to see if the step result for this profile option already exists
+            StepResult profileResult = StepResultHelper.findStepResult(result, option.getIdentifier());
+            // If it doesn't exist, create one matching the profile info option type from User object
+            if (profileResult == null) {
+                Step profileStep = StepHelper.getStepWithIdentifier(formStep.getFormSteps(), option.getIdentifier());
```
liujoshua: I'm a bit confused. A FormStep is a Step which itself contains Steps.  What makes FormStep different from (or, why does it not extend) Task, which can also contain multiple steps?

backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItem.java
liujoshua: Maybe we can use this https://developer.android.com/reference/java/util/function/Predicate.html

And have a util class that parses a string and returns a Predicate?

backbone/src/main/java/org/researchstack/backbone/model/survey/SubtaskQuestionSurveyItem.java
```diff
@@ -10,27 +10,4 @@
     SubtaskQuestionSurveyItem() {
```
liujoshua: I haven't had a chance to look at Survey in iOS and how it relates to Question and (sub)Task, but this name makes it sounds complex!

liujoshua:
Looks good to me

backbone/src/main/java/org/researchstack/backbone/model/User.java
```diff
@@ -68,9 +68,9 @@ public void setDataSharingScope(DataSharingScope dataSharingScope) {
      * ALL   - The user has consented to sharing their de-identified data for current and future research, which may or may not involve the same institutions or investigators.
      */
     public enum DataSharingScope {
-        NONE("none"),
-        STUDY("study"),
-        ALL("all");
+        NONE("no_sharing"),
```
liujoshua: Nice. Want to replace literals with this enum and also fix this in your PR? https://sagebionetworks.jira.com/browse/MA-48

backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItem.java
dephillipsmichael: Predicate class requires API level 24 or higher.  But also, it really isn't necessary, iOS was using it in an odd way.  The way I did NavigableSteps I think is more understandable anyways.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ProfileStepLayout.java
dephillipsmichael: A FormStep displays multiple QuestionSteps in one StepLayout screen.  For instance, Eligibility Quiz is a FormStep. Tasks are meant to simply control the movement from step to step, so if you put a list of steps in a Task, it would put each QuestionStep on their own screen, where a FormStep needs multiple steps on one screen.

backbone/src/main/java/org/researchstack/backbone/model/survey/SubtaskQuestionSurveyItem.java
dephillipsmichael: Survey only relates to the JSON model.  So this one is a SubtaskStep that contains only questions.  Consent Quiz is a SubtaskQuestionSurveyItem.

backbone/src/main/java/org/researchstack/backbone/model/User.java
dephillipsmichael: Good look on this.  I think it is fixed now.

dephillipsmichael:
Need to fix unit tests

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ConsentReviewSubstepListStepLayout.java
dephillipsmichael: @liujoshua I ended up fixing this, see StepLayoutHelper method safePerform, and how it's used in LoginStepLayout.

PR #323, opened 2017-01-24
Maven deploy plugin, change to 2.0.0-SNAPSHOT

liujoshua:
This allows you to run './gradlew install' and publish the artifact to the local maven repo


liujoshua:
This changes the artifact version to 2.0.0-SNAPSHOT on the develop branch. 

The next release (merge to master) will be a major version change since it won't be backwards compatible, hence the jump to 2.0.0. 

"-SNAPSHOT" is used in Maven to indicate a version that is under heavy development, shouldn't be cached long, etc. 

When a release branch is formed, -SNAPSHOT can be dropped in lieu of other suffices like 2.0.0-rc1 if there is a need to let the code bake. The final release and merge to master will then be versioned as 2.0.0

PR #324, opened 2017-01-25
Feature/change email

dephillipsmichael:
Added the change email screen, which can be accessed from the EmailVerification screen.  To accomplish this, and keep the password secure, I put the EmailVerificationSubstep and a substep of RegistrationStep together in a view pager so they could communicate with each other through their owning class which is EmailVerificationStepLayout.

In this PR, I also added support for ViewPagerSubstepLayouts saving their index and reloading them if the view was created/destroyed.


backbone/src/main/java/org/researchstack/backbone/model/survey/factory/SurveyFactory.java
```diff
@@ -168,7 +175,7 @@ public Step createSurveyStep(Context context, SurveyItem item, boolean isSubtask
                 if (!(item instanceof InstructionSurveyItem)) {
                     throw new IllegalStateException("Error in json parsing, ACCOUNT_EMAIL_VERIFICATION types must be InstructionSurveyItem");
                 }
-                return createEmailVerificationStep((InstructionSurveyItem)item);
+                return createEmailVerificationStep(context, (InstructionSurveyItem)item);
```
liujoshua: General comment not just on your PR: I'd like to propose clarifying which context is expected, to make it easier for consumers of ResearchStack to call correctly, and also make it easier for us to code correctly.

For instance, we've removed a lot of places where applicationContext is being passed, which is awesome since it's completely unnecessary. 

But in a lot of places, we actually require an activityContext and the method signature doesn't make that clear.

backbone/src/main/java/org/researchstack/backbone/model/survey/factory/SurveyFactory.java
```diff
@@ -632,12 +639,21 @@ public LoginStep createLoginStep(Context context, ProfileSurveyItem item) {
...
     /**
      * @param item InstructionSurveyItem from JSON
-     * @return valid EmailVerificationStep matching the InstructionSurveyItem
+     * @return valid EmailVerificationSubStep matching the InstructionSurveyItem
      */
-    public EmailVerificationStep createEmailVerificationStep(InstructionSurveyItem item) {
-        EmailVerificationStep step = new EmailVerificationStep(item.identifier, item.title, item.text);
-        fillInstructionStep(step, item);
-        return step;
+    public EmailVerificationStep createEmailVerificationStep(Context context, InstructionSurveyItem item) {
```
liujoshua: Nit-pick: change method signature return type to EmailVerificationSubStep to match the change to the comment

backbone/src/main/java/org/researchstack/backbone/model/survey/factory/SurveyFactory.java
dephillipsmichael: Good point.  This can be any context, just used for string resources.  I think I mention that in the header of constructor of SurveyFactory.

PR #332, opened 2017-01-27
Feature/share the app

dephillipsmichael:
Added share the app step and layout.  In Mpower, this shows after you are ineligible.  Any other app will now be able to add it to their on boarding steps with JSON as well.


backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ShareTheAppStepLayout.java
```diff
@@ -0,0 +1,311 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import rx.functions.Action1;
+
+/**
+ * Created by TheMDP on 1/26/17.
+ */
+
+public class ShareTheAppStepLayout extends FixedSubmitBarLayout implements StepLayout {
```
dephillipsmichael, line 38: Most of this class was copied from ShareFragment

skin/src/main/java/org/researchstack/skin/task/SignUpTask.java
```diff
@@ -57,60 +57,60 @@ private void initSteps(Context context) {
...
-                    {
-                        // TODO: extend the json to include (yes/no)?
-                        BooleanAnswerFormat booleanAnswerFormat =
-                                new BooleanAnswerFormat(context.getString(R.string.rsb_yes), context.getString(R.string.rsb_no));
-                        for (InclusionCriteriaModel.Item item : s.items) {
-                            QuestionStep question = new QuestionStep(item.identifier, item.text, booleanAnswerFormat);
-                            answerMap.put(item.identifier, item.expectedAnswer);
-                            questions.add(question);
+            // step can be null with addition of new OnboardingManager steps, so just ignore them
+            if (s != null && s.type != null) {
```
dephillipsmichael, line 61: stupid git diff... all I did was add the conditional line, the rest is indentation.

backbone/src/main/java/org/researchstack/backbone/utils/ResUtils.java
```diff
@@ -13,6 +13,12 @@
 
 public class ResUtils
 {
+    public static final String LOGO_DISEASE = "logo_disease";
```
liujoshua, line 16: should this have the rsb_ prefix?

liujoshua (approved):
Looks good, one minor question

rianhouston:
Code looks good.  The ShareFragment was available from the Learn screen as well.  Does this replace that?

backbone/src/main/java/org/researchstack/backbone/utils/ResUtils.java
dephillipsmichael: Well, this resource is a little different than the patterns of the rest.  Most of the resources in backbone have a default value, which in this case would be a default image, that shows up if the app doesn't override that resource name.  However, this one just assumes all apps will override it with their own and it will be named "logo_disease".  So, that is why it doesn't have "rsb_" and the compiler won't complain.

dephillipsmichael:
@rianhouston yeah, this will eventually replace that as well, the other one is now deprecated.  I'll make a story for it.

PR #333, opened 2017-01-27
Feature/completion steps

dephillipsmichael:
Added image support for InstructionStepLayout.
These images can also be AnimatedVectors, if isImageAnimated is true.
Used these two additions to make the CompletionStepLayout, which is used in Onboarding after the user passes the consent quiz. 

To test this in MPower, see https://github.com/Sage-Bionetworks/mPower-Android/pull/21


EricSiegNW:
I think this is related to the previous PR, but I'm getting an error when trying to sync gradle atm:

In rsb_step_layout_share_the_app.xml it fails at the android:src="@drawable/icon" line in the rsb_share_logo_view ImageView stating that no resource was found that matches the given name (at 'src' with value '@drawable/icon')

This is in spite of me being on latest mPower meant to test the changes you have here.

Any idea what might be causing this?

PR #336, opened 2017-01-29
Bug/build warnings

dephillipsmichael:
Fixed build warnings.  Oddly enough, they all came from Javadoc, which makes me think the culprit here is the new Java 8 doclint task.  Also, I was unable to solve the R.Java errors that also stem from the javadoc being built.  Try tons of approaches, like exclude R.java, adding R to class path, and even disabling doclint, but literally nothing worked.

### DIFF
--- a/backbone/build.gradle
+++ b/backbone/build.gradle
@@ -35,8 +35,7 @@ android {
         exclude 'META-INF/NOTICE.txt'
     }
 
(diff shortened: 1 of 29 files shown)


liujoshua (approved):
Thanks for fixing all these errors and eliminating the noise!

PR #337, opened 2017-01-30
Merge master to develop

liujoshua:
A few changes in master aren't in develop yet.

Giant diff because one of those changes was applying different formatter


dephillipsmichael:
I tried my best to give it the once over, I think everything looks ok.

PR #338, opened 2017-01-31
Feature/update dependencies

liujoshua:
- Performs `./gradlew build` instead of `./gradlew test`, so errors which show up in build but not test will be caught here, instead of packages which pull this repo in as a submodule
- Reduce noise from tools breaking on retrolambda/lombok
- Reduce javadoc noise on `:backbone`, unsure why same code does not work for `:skin` and why `:skin` is not including generated classes the same as backbone
- Updates dependencies. Currently, BridgeAndroidSDK and mPower will substitute in newer versions via dependency resolution, but depending on which project is the root, the result can vary greatly. Updated to be closer in version. This was especially annoying when Studio would pick compile against one version, but its tooling would use another version (not sure why or how this happens)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ android:
     # The BuildTools version used by your project
     # per travis-ci issue #5036, add '-tools' to use build-tools-23.0.2
     - tools
-    - build-tools-25.0.0
+    - build-tools-25.0.2
 
     # The SDK version used to compile your project
     - android-25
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -31,12 +31,7 @@ before_install:
   - export JAVA8_HOME=/usr/lib/jvm/java-8-oracle
   - export JAVA_HOME=$JAVA8_HOME
 
-before_script:
-  # Tests crash if a local.properties file isn't present; this file is not kept
-  # under version control. Create an empty local.properties for testing.
-  - touch local.properties
-
-script: ./gradlew test
+script: ./gradlew build
```
dephillipsmichael, line 34: Good idea

dephillipsmichael:
Approved

PR #341, opened 2017-02-03
Feature/fingerprint keystore

dephillipsmichael:
I did a full security review on the current encryption functionality of the SDK.  The default 4-digit pin code is the least secure method of encryption, the password is more secure (but at that point we should just use login credentials) and the most secure and convenient method, fingerprint, was not supported. 

I added the fingerprint functionality, enforced by default if the phone supports it.   Also, FYI, you can use the emulator to test it see http://stackoverflow.com/questions/35335892/android-m-fingerprint-scanner-on-android-emulator just remember to do the telnet command and follow the first answer

For QA and reviewing there are a few test cases that should be run:
1) A passcode/password should still be used if the device does not support fingerprints, or if the device does not have any fingerprints registered.
2) If the user removes their fingerprints after it is registered with the SDK, they will be logged out, and they must go through the process again
3) If user adds another fingerprint after it is registered with the SDK, they will be logged out, and they must go through the process again
4) if the user registers a fingerprint for the first time, and then navigates back to creation screen, it should be skipped


backbone/src/main/java/org/researchstack/backbone/ui/step/layout/FingerprintStepLayout.java
```diff
@@ -0,0 +1,479 @@
...
+        if (!(step instanceof FingerprintStep)) {
+            throw new IllegalStateException("FingerprintStepLayout expects a FingerprintStep");
+        }
+
+        fingerprintStep = (FingerprintStep)step;
+    }
+
+    protected FingerprintManagerCompat.AuthenticationCallback authCallbacks = new FingerprintManagerCompat.AuthenticationCallback() {
+
+        static final int ERROR_MSG_TOO_MANY_ATTEMPTS = 7;
```
rianhouston: Is this something that individual apps may want to set and should be passed in somehow?

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/FingerprintStepLayout.java
```diff
@@ -0,0 +1,479 @@
...
+ * Created by TheMDP on 1/31/17.
+ *
+ * This class can be used as a more convenient and secure method for creating
+ * an encryption key that will be used to protect the user's data
+ *
+ * It works with the FingerprintManager to create a secret key that is backed by the user's fingerprint
+ */
+
+@TargetApi(android.os.Build.VERSION_CODES.M) // api 23, or Android 6.0
+public class FingerprintStepLayout extends InstructionStepLayout {
```
rianhouston, line 38: I think this is the convention but it confuses me to put all this business logic in a class called 'Layout'.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/FingerprintStepLayout.java
```diff
@@ -0,0 +1,479 @@
...
+
+    /**
+     * Initialize the {@link Cipher} instance with the created key in the {@link #createKey()}
+     * method.
+     *
+     * @return {@code true} if initialization is successful, {@code false} if the lock screen has
+     * been disabled or reset after the key was generated, or if a fingerprint got enrolled after
+     * the key was generated, which in all cases we must ask the user to sign back in with their DataProvider credentials
+     */
+    private boolean initCipher() {
```
rianhouston: Anyway to separate some of this keystore/encryption logic into a utility class?

rianhouston:
This looks like a very cool feature.  Code looks good, just a few questions/comments.  I haven't had a chance to exercise it yet though.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/FingerprintStepLayout.java
dephillipsmichael: I added a comment here, and renamed it, because this is actually an error code.  The hardware shuts off after "X" failed attempts, and we show the error message and boot the user from the screen in that case.

dephillipsmichael: That did motivate me to move the animation delay times to R.integer though so they can be customized.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/FingerprintStepLayout.java
dephillipsmichael: No, you are right.  The keystore/fingerprint/encryption code should go in some sort of security utils class.  I'll do that.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/FingerprintStepLayout.java
dephillipsmichael: No, you are right. The keystore/fingerprint/encryption code should go in some sort of security utils class. I'll do that.

dukemike:
First of all, thanks for implementing this.

Should the user be _required_ to use their fingerprint? One nice thing about RK's implementation which makes it optional is that my wife can use my phone to use the app (granted, this scenario isn't viable certain types of studies.) Also, since the device itself allows a passcode as a fallback, it would be nice if falling back was supported as well. I'd prefer for this to be a configurable item, with the default being the most secure (fingerprint only as you've described.)

dephillipsmichael:
@dukemike thanks for taking a look at this.  You bring up some good points, and I thought about them quite a bit before I started implementing any of this code.  

My first priority is always implementing what iOS has.  Which in this case, like you mentioned, is have the fingerprint be optional.  Unfortunately, if you did this, it would remove the security improvements that the fingerprint has over the pin code. This is because the encrypted secret you save will have to be the pin code, and that would still only be a 4-digit pin code.  Where as with the fingerprint, it is a securely random string with 32 characters, which is much more secure and infinitely harder to crack.  

Also, in response to your concern at a fundamental UX level, Android allows you to add as many fingerprints as you want, so your wife could simply add her fingerprint to your device, and the app would authenticate her fine.  The code works with any of your registered fingerprints in the OS, not just the one you initially use.  

I hope that addresses your concern. 

Another option down the road would be to just use the device’s pin/pattern/password/fingerprint with the ConfirmCredential alert available in api > 21 https://developer.android.com/samples/FingerprintDialog/index.html which does what you are referring to as well.  However, I didn’t go with this as it is only available in api > 21 and we already have this functionality basically built into PincodeLayout

dephillipsmichael:
@dukemike thanks for taking a look at this.  You bring up some good points, and I thought about them quite a bit before I started implementing any of this code.  

My first priority is always implementing what iOS has.  Which in this case, like you mentioned, is have the fingerprint be optional.  Unfortunately, if we did this, it would remove the security improvements that the fingerprint has over the pin code. This is because the encrypted secret you would save would have to be the pin code, and that would still only be 4 numeric digits.  Where as with the fingerprint, it is a securely random string with 32 characters, which is more secure and infinitely harder to crack.  

Also, in response to your concern at a fundamental UX level, Android allows you to add as many fingerprints as you want, so your wife could simply add her fingerprint to your device, and the app would authenticate her fine.  The code works with any of your registered fingerprints in the OS, not just the one you initially use.  

I hope that addresses your concerns, I'm certainly open to discussing it furthur.

dukemike:
@dephillipsmichael I must run into more problems with the finger print reader than you :) I like the  feature but prefer it to be optional. I'm currently working on a CareKit app which contains a lot of educational materials in addition to symptom tracking, and in this scenario I'd prefer to allow a care giver to easily use the app on the patient's device as well. Of course, the patient could unlock the app for them, but giving them the pin code is easier, especially if the device is set to lock fairly quickly or it becomes a pain to ask them to unlock their phone repeatedly.
 
I completely understand where you're coming from though.

dephillipsmichael:
@dukemike I do see where you're coming from.  It really should be optional if for any reason the user's conditions require the passcode.  It shouldn't be too difficult to implement, but it still won't operate like iOS exactly.  The user will have the option to just use passcode exclusively or fingerprint exclusively, with the ability to switch between those auth methods like "forgot passcode" functions.

dephillipsmichael:
@dukemike @EricSiegNW on the fingerprint screen I added a "use passcode" in the submit bar with a warning alert saying it is less secure but can be useful.  It switches to passcode layout and lets a user set a passcode.

liujoshua (approved):
Win for UX and security 👍

dephillipsmichael:
Merging :)

PR #343, opened 2017-02-05
Feature/localization

dephillipsmichael:
I'm not sure why this was never done, but this will be really helpful as we start making all the active tasks.

### DIFF
--- a/backbone/src/main/res/values-ar/strings.xml
+++ b/backbone/src/main/res/values-ar/strings.xml
@@ -1,0 +1,396 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <!-- Informed consent process. -->
+    <string name="rsb_CONSENT_NAME_TITLE">موافقة</string>
+    <string name="rsb_CONSENT_NAME_GIVEN">الاسم الأول</string>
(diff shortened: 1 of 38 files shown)


liujoshua:
This is great. One idea is moving these to a new resource file to keep them separated instead of strings.xml, perhaps something like researchkit-generated-strings.xml?

liujoshua (approved):

dephillipsmichael:
I'm not familiar on how you would do that.  I thought they had to be named "strings.xml"

pthealthtech:
Was this implemented? We are trying to deliver surveys in China and the step instructions are not translating.

dephillipsmichael:
@pthealthtech this is available in the develop branch, but the develop branch is not ready for release.  I'd suggest checking out the develop branch locally, and copy/pasting the strings.xml and the corresponding languages you need.  

However, note that this just makes available the strings, it doesn't automatically include them in the created steps. You will need to provide separate language translations for "onboarding.json", "consent.json", and "eligibility_requirements.json" and any and all corresponding files you make tasks out of, and create each depending on the device's current Locale.  This is a feature that is not implemented on the develop branch either.  We don't have any apps that are localized in anything but English ATM.

PR #344, opened 2017-02-06
Feature/tremor task

dephillipsmichael:
This PR contains an implementation of the Tremor Task.  It is created through the OrderedTaskFactory, which will be where all the future Activities are created.

This code also lays the foundation for all future Tasks.  It adds a base class SensorRecorder that can record any Android sensor data at specific frequency, and asynchronously write the data as a JSON array to a file.  This PR has implemented the AccelerometerRecorder and the DeviceMotionRecorder which are what is recorder in the Tremor Task.

Also introduced is the DataLoggerManager.  This class is central station for all files created by Records.  It manages deleting "dirty" files, or files that were reordered but never reached the point of attempting to upload.  It manages and keeps files around that have been complete, but not successfully uploaded yet, and once the SDK manager for upload is integrated, it will manage removing ones that have been successfully uploaded.

This also adds some new StepLayouts.  These include the CountdownStepLayout, and the ActiveStepLayout.  The ActiveStepLayout is the most important as it is the brains behind managing the Recorder state.  As of now, there is only start and stop functionality, with no pause/resume functionality.  I'm not sure if that is needed or not yet.  To avoid those potential scenarios, the ActiveStepLayout locks the screen rotation, and also keeps the screen on.  Please test the life cycle of the ActiveStepLayout while reviewing this PR.

The creation of the TremorTask is also finished, and all scenarios of it's creation, including multiple options, are unit tested.


backbone/src/main/java/org/researchstack/backbone/task/OrderedTaskFactory.java
```diff
@@ -0,0 +1,643 @@
...
+import java.util.Random;
+
+/**
+ * Created by TheMDP on 2/4/17.
+ *
+ * In iOS, they included a bunch of static methods for building OrderedTasks in the
+ * OrderedTask class.  However, I think they belong in this Factory class
+ */
+
+public class OrderedTaskFactory {
```
dephillipsmichael: Not sure why this class wasn't showing up in my diff online, commenting here so you guys see it.

rianhouston:
This code looks good to me for a WIP.

dephillipsmichael:
@rianhouston this code was updated to be the finished product, so it is worth another look.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ActiveStepLayout.java
```diff
@@ -0,0 +1,463 @@
...
+ * for the task.
+ *
+ * When you develop a new active step, you should subclass `ActiveStepLayout`
+ * and define your specific UI. When subclassing, pay special attention to the life cycle
+ * methods, `start`, `finish`, `suspend`, and `resume`. Also, be sure to test for
+ * the expected behavior when the user suspends and resumes the app, during task
+ * save and restore, and during UIKit's UI state restoration.
+ */
+
+public class ActiveStepLayout extends FixedSubmitBarLayout implements StepLayout, RecorderListener, TextToSpeech.OnInitListener {
```
dephillipsmichael: I'm not sure why this diff wasn't showing, but please review this class.

backbone/src/main/AndroidManifest.xml
```diff
@@ -4,5 +4,7 @@
 
     <!-- We allow users to secure their data using their fingerprint see -->
     <uses-permission android:name="android.permission.USE_FINGERPRINT"/>
+    <!-- ActiveSteps may vibrate the phone on start and on finish -->
+    <uses-permission android:name="android.permission.VIBRATE"/>
```
rianhouston: Should this be the responsibility of the apps rather than backbone?  Not all apps would necessarily use ActiveSteps right?  Does MoleMapper?

dephillipsmichael: That's a good point.  I'll comment it out and write that it is the responsibility of the app to include.

backbone/src/main/java/org/researchstack/backbone/result/logger/DataLoggerManager.java
```diff
@@ -0,0 +1,207 @@
...
+ */
+
+public class DataLoggerManager {
+
+    /**
+     * SharedPreferences are used to keep track of the DataLogger files' status
+     * They can be dirty while being written, completed, and uploaded
+     */
+    private SharedPreferences sharedPrefs;
+    private static final String SHARED_PREFS_KEY = "DataLoggerManagerSharedPrefs";
```
rianhouston, line 37: Interesting use of SharedPrefs here.  I think its fine.  I have done similar things by writing a second file to track status.  That could eliminate the dependency on the android context and allow you test more easily.

backbone/src/main/java/org/researchstack/backbone/step/active/AccelerometerRecorder.java
```diff
@@ -0,0 +1,75 @@
...
+            jsonObject = new JsonObject();
+        }
+    }
+
+    /**
+     * Called by the base class at the Recorder's frequency
+     */
+    @Override
+    protected void writeJsonData() {
+        jsonObject.addProperty(TIMESTAMP_KEY, System.currentTimeMillis());
```
rianhouston, line 58: Are we appending to the jsonObject and then rewriting the whole object each time?  Maybe it will become more obvious as I read further.

backbone/src/main/java/org/researchstack/backbone/step/active/ActiveStep.java
```diff
@@ -0,0 +1,338 @@
...
+     */
+    private String finishedSpokenInstruction;
+
+    /**
+     * An image to be displayed below the instructions for the step.
+     *
+     * The image can be stretched to fit the available space. When choosing a size
+     * for this asset, be sure to take into account the variations in device form factors.
+     */
+    private String image;
```
rianhouston: What is the string?  Does this map to a string resource identifier?  I think the getter/setter may be confusion here.  getImage() returns a string?   Maybe rename to what it acutally is?

backbone/src/main/java/org/researchstack/backbone/step/active/DeviceMotionRecorder.java
```diff
@@ -0,0 +1,136 @@
...
+                Sensor.TYPE_LINEAR_ACCELERATION,
+                Sensor.TYPE_GYROSCOPE,
+                Sensor.TYPE_MAGNETIC_FIELD,
+                Sensor.TYPE_ROTATION_VECTOR
+        );
+    }
+
+    @Override
+    public void onSensorChanged(SensorEvent sensorEvent) {
+        if (sensorEvent.sensor.getType() == Sensor.TYPE_ACCELEROMETER) {
```
rianhouston: A switch statement may be better here.

backbone/src/main/java/org/researchstack/backbone/step/active/DeviceMotionRecorderConfig.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.File;
+
+/**
+ * Created by TheMDP on 2/5/17.
+ */
+
+public class DeviceMotionRecorderConfig extends RecorderConfig {
+
+    /**
+     * The frequency of accelerometer data collection in samples per second (Hz).
```
rianhouston: Is this comment out of date?  Its more than the accelerometer right?

backbone/src/main/java/org/researchstack/backbone/step/active/AccelerometerRecorder.java
dephillipsmichael: The "addProperty" method adds or overwrites the the current property. So this just sets the value.  Then, we write the whole json object to file in the overridden "writeJsonData()" method.

backbone/src/main/java/org/researchstack/backbone/result/logger/DataLoggerManager.java
dephillipsmichael: That's a valid point.  I think you can mock SharedPreferences fairly easily with Mockito.  However, this brings up the point that DataLoggerManager should be unit tested.

backbone/src/main/java/org/researchstack/backbone/task/OrderedTaskFactory.java
```diff
@@ -0,0 +1,621 @@
...
+            List<TaskExcludeOption> taskOptionList)
+    {
+        // Coin toss for which hand first (in case we're doing both)
+        final boolean leftFirstIfDoingBoth = (new Random()).nextBoolean();
+        return tremorTask(context, identifier, intendedUseDescription, activeStepDuration,
+                tremorOptionList, handOption, taskOptionList, leftFirstIfDoingBoth);
+    }
+
+    // This method is separate mainly for unit testing purposes, to eliminate randomness
+    protected static NavigableOrderedTask tremorTask(
```
rianhouston: Is this task always the same?  Aren't some tasks defined by service responses?  Or is that only for surveys?

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ActiveStepLayout.java
```diff
@@ -0,0 +1,463 @@
...
+
+    private static final int DEFAULT_VIBRATION_AND_SOUND_DURATION = 500; // in milliseconds
+
+    /**
+     * When this is true, files will be saved externally so you can read them
+     * Reading internal files requires root access
+     * You must add WRITE_EXTERNAL_STORAGE permission to manifest as well
+     */
+    private static final boolean DEBUG_SAVE_FILES_EXTERNALLY = false;
```
rianhouston: Nice!

backbone/src/main/java/org/researchstack/backbone/step/active/DeviceMotionRecorderConfig.java
dephillipsmichael: fixed

backbone/src/main/java/org/researchstack/backbone/step/active/DeviceMotionRecorder.java
dephillipsmichael: fixed

backbone/src/main/java/org/researchstack/backbone/step/active/ActiveStep.java
dephillipsmichael: i ended up removing "Image" from ActiveStep, doesn't look like iOS was even using it

backbone/src/main/java/org/researchstack/backbone/task/OrderedTaskFactory.java
dephillipsmichael: I'm not sure I understand the question.  Is it the same task in which regard?

backbone/src/test/java/org/researchstack/backbone/task/TremorTaskTest.java
```diff
@@ -0,0 +1,280 @@
...
+    private Resources mockResources;
+
+    @Before
+    public void setUp() throws Exception {
+        mockContext = Mockito.mock(Context.class);
+        mockResources = Mockito.mock(Resources.class);
+        Mockito.when(mockContext.getResources()).thenReturn(mockResources);
+
+        // All the strings that the TremorTask uses
+        Mockito.when(mockContext.getString(R.string.rsb_TREMOR_TEST_ACTIVE_STEP_FINISHED_INSTRUCTION)).thenReturn("");
```
rianhouston, line 41: Oh my!

rianhouston:
This looks great and is extremely well documented.  I learned a lot by reading the code and the javadocs!

rianhouston:
:shipit:

backbone/src/main/java/org/researchstack/backbone/task/OrderedTaskFactory.java
rianhouston: I meant does a TremorTask always have the same steps.  But I see that you built a way to exclude certain steps so I think that answers my question and you can ignore this comment.

backbone/src/main/java/org/researchstack/backbone/step/active/AccelerometerRecorder.java
rianhouston: Ah, I missed that nuance.  Sounds good.

backbone/src/main/java/org/researchstack/backbone/task/OrderedTaskFactory.java
liujoshua: Looks like so far it's just TremorTask, and this file is already 600 lines, and I'm guessing each task will take hundreds of lines to create. Can we call this TremorTaskFactory? If there's a need for one class for all OrderedTask, OrderedTaskFactory could call just call *TaskFactory. Just so these files are smaller.

EDIT by Mike: good point, i renamed it TremorTaskFactory

backbone/src/main/java/org/researchstack/backbone/step/active/SensorRecorder.java
```diff
@@ -0,0 +1,148 @@
...
+    private int  writeCounter;
+    private long writeDelayGoal;
+    private long writeStartTime;
+
+    /** Default constructor for serialization/deserialization */
+    SensorRecorder() {
+        super();
+    }
+
+    SensorRecorder(double frequency, String identifier, Step step, File outputDirectory) {
```
liujoshua, line 54: Random thought here, do we want to define a way accommodate non time-series sensors? Weight, blood pressure.

backbone/src/test/java/org/researchstack/backbone/task/TremorTaskTest.java
liujoshua: LOL

backbone/src/main/res/drawable/rsb_tremortest1.xml
```diff
@@ -0,0 +1,88 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
liujoshua: Nit-pick: Having test in the file name is slightly confusing. If there's another name that makes sense, I'd prefer that.

backbone/src/main/java/org/researchstack/backbone/utils/ResUtils.java
```diff
@@ -18,6 +18,21 @@
     public static final String ERROR_ICON                   = "rsb_error";
     public static final String IC_FINGERPRINT               = "rsb_fingerprint";
 
+    public static final String TREMOR_TEST_1                = "rsb_tremortest1";
```
liujoshua: This suggestion totally goes against the rest of the code so take it for what it's worth... I like grouping constants with nested inner classes, So I can do something like ResUtils.Tremor.TEST_1.. or static import ResUtils.Tremor and have Studio auto-completing the constants relevant to Tremor.

public static class Tremor {
  public static final String TEST_1 = "rsb_tremortest1";
  public static final String TEST_2 = "rsb_tremortest2";
}

backbone/src/main/java/org/researchstack/backbone/result/logger/DataLoggerManager.java
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * This will loop through all the data logger files that exist in storage and
+     * remove any that are "dirty".  A "dirty" file is any file that has not been zipped
+     * up and attempted to be uploaded yet.
+     *
+     * Examples of dirty data logger files are ones that were written before and app crashed
+     * or was force closed, or files that were written and then the user cancelled the active task
+     * that was responsible for creating them
+     */
+    public void deteleAllDirtyFiles() {
```
liujoshua: typo

liujoshua (commented):
First active task! Looking good to me

backbone/src/main/java/org/researchstack/backbone/step/active/SensorRecorder.java
dephillipsmichael: The Recorders only exist for results that are too large to save in app memory, so non-time-series sensors don't really apply to the paradigm of SensorRecorder.  I think we would just make a separate StepLayout to accommodate and pull weight, blood pressure, etc.

backbone/src/main/res/drawable/rsb_tremortest1.xml
dephillipsmichael: Yeah, I can totally see that.  I was just copying iOS.  I'll rename them to something more like you suggested.

PR #345, opened 2017-02-06
Update retrofit version to match android-sdk

rianhouston:
This mismatch in retrofit versions was causing errors in studio for me.


dephillipsmichael:
approved

PR #348, opened 2017-02-11
Feature/updates for new consent flow

dephillipsmichael:


skin/src/main/java/org/researchstack/skin/ui/OverviewActivity.java
```diff
@@ -138,7 +138,9 @@ public void onDataAuth() {
     @Override
     public void onResume() {
         super.onResume();
-        if (DataProvider.getInstance().isSignedIn(this)) {
+        // We shouldn't be able to get here if the user is signed in and consented,
+        // Go to MainActivity instead
+        if (DataProvider.getInstance().isSignedIn(this) && DataProvider.getInstance().isConsented()) {
```
dephillipsmichael, line 143: This was the most important change.

liujoshua: Ahh yes. I remember us talking about the ambiguity of isSignedIn... I totally didn't mean to change the logic! But this does make more sense to me. They're signedIn in the sense they have a session and we know their email/password... we're just getting a 412 on signIn and other APIs, but not a 403

liujoshua (approved):
Looks good

PR #349, opened 2017-02-17
Feature/ma 72

dephillipsmichael:
This PR adds a fully functional short walk task and tests.  

The short walk test asks the user to walk for X number of steps, while they do this, the device will  be recording the number of steps the user takes using the pedometer sensor if available, or if not available, an accelerometer pedometer fusion algorithm I found on the internet.  I think it needs a few iterations of improvement, so please offer suggestions if you have any.  Anyways, while this is happening, it also records accelerometer and device motion values.  The step goes for either X number of steps, or a timeout in seconds, based on number of steps.  There are two steps like this, and then a resting step in which the user stays still.  

This also includes the walk back and forth task.


backbone/src/main/java/org/researchstack/backbone/step/active/Recorder.java
```diff
@@ -117,6 +117,15 @@
...
     public abstract void stop();
 
+    /**
+     * A force stop will cause this recorder to be immediately cancelled,
+     * the file it was writing will be deleted,
+     *
+     * and no callback will be invoked
+     */
+    @MainThread
+    public abstract void forceStop();
```
liujoshua: DataLogger has a method cancel() that sounds like it has the same behavior. Could we name them the same?

liujoshua (approved):

backbone/src/main/java/org/researchstack/backbone/result/logger/DataLoggerFileWriterThread.java
```diff
@@ -178,6 +179,10 @@ private void writeCanceledFromThreadToMainThread() {
             @Override
             public void run() {
                 thread.quit();
+                boolean success = file.delete();
+                if (!success) {
+                    Log.d(getClass().getSimpleName(), "Failed to delete file " + file.toString());
```
rianhouston: We seem to have a lot of inconsistency around logging.  Backbone has a LogExt class that is used elsewhere.  I don't really have a preference, but consistency would be nice.

backbone/src/main/java/org/researchstack/backbone/step/RequireSystemFeatureStep.java
```diff
@@ -0,0 +1,45 @@
+package org.researchstack.backbone.step;
+
+import org.researchstack.backbone.ui.step.layout.RequireSystemFeatureStepLayout;
+
+/**
+ * Created by TheMDP on 2/21/17.
+ *
+ * This step forces the user to turn on a system feature before proceeding in the task
+ * for not it is just GPS, but this could be expanded to cover Wifi, Cellular, NFC, etc
```
rianhouston: Typo:  not -> now?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/LocationRecorder.java
```diff
@@ -0,0 +1,153 @@
...
+            locationManager.requestLocationUpdates(
+                    LocationManager.NETWORK_PROVIDER, minTime, minDistance, this);
+        } catch (java.lang.SecurityException ex) {
+            Log.i(TAG, "fail to request location update, ignore", ex);
+        } catch (IllegalArgumentException ex) {
+            Log.d(TAG, "network provider does not exist, " + ex.getMessage());
+        }
+
+        try {
+            locationManager.requestLocationUpdates(
```
rianhouston: Do you really want to request updates from both NETWORK and GPS?  As opposed to trying GPS but fallback to NETWORK or such?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/LocationRecorder.java
```diff
@@ -0,0 +1,153 @@
...
+    @Override
+    public void onLocationChanged(Location location) {
+        if (location != null) {
+            jsonObject.addProperty(TIMESTAMP_KEY, location.getTime());
+
+            coordinateJsonObject.addProperty(LONGITUDE_KEY, location.getLongitude());
+            coordinateJsonObject.addProperty(LATITUDE_KEY, location.getLatitude());
+            jsonObject.add(COORDINATE_KEY, coordinateJsonObject);
+
+            if (location.hasAccuracy()) {
```
rianhouston, line 136: Is it worth rejecting updates entirely if the accuracy is > a threshold?  Or I suppose that could be done in the data analysis.

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/FitnessStepLayout.java
```diff
@@ -0,0 +1,48 @@
...
+import android.content.Context;
+import android.util.AttributeSet;
+
+import org.researchstack.backbone.step.Step;
+import org.researchstack.backbone.step.active.FitnessStep;
+
+/**
+ * Created by TheMDP on 2/16/17.
+ *
+ * This exists in iOS and the class is used to monitor certain special Recrods,
```
rianhouston: Typo?

backbone/src/main/java/org/researchstack/backbone/utils/FormatHelper.java
```diff
@@ -8,6 +11,12 @@
 
 public class FormatHelper {
 
+    private static final String COUNTRY_US      = "US";
+    private static final String COUNTRY_LIBERIA = "LR";
```
rianhouston, line 15: Why these 3 countries?

backbone/src/main/res/layout/rsb_step_layout_active_step.xml
```diff
@@ -25,4 +25,22 @@
...
+        style="@style/Base.Widget.AppCompat.ProgressBar"
+        android:id="@+id/rsb_active_step_layout_progress"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:padding="@dimen/rsb_padding_large"
+        android:indeterminate="true"
+        android:visibility="gone"/>
+
+    <android.support.v7.widget.AppCompatImageView
```
rianhouston: Do you need to specify this a AppCompatImageView?  In the docs it says it will automatically be used if you specify ImageView:  https://developer.android.com/reference/android/support/v7/widget/AppCompatImageView.html

backbone/src/test/java/org/researchstack/backbone/task/factory/WalkingTaskTests.java
```diff
@@ -0,0 +1,486 @@
...
+    private Resources mockResources;
+
+    @Before
+    public void setUp() throws Exception {
+        mockContext = Mockito.mock(Context.class);
+        mockResources = Mockito.mock(Resources.class);
+
+        Mockito.when(mockContext.getPackageName()).thenReturn(PACKAGE_NAME);
+
+        mockLocationManager = Mockito.mock(LocationManager.class);
```
rianhouston, line 61: Can't you use the @Mock annotation to setup your mocks?

rianhouston:
Looks pretty good to me.  Left a few comments/questions.

backbone/src/main/java/org/researchstack/backbone/step/active/Recorder.java
dephillipsmichael: Done

backbone/src/main/java/org/researchstack/backbone/result/logger/DataLoggerFileWriterThread.java
dephillipsmichael: Good point, I went ahead and changed it to LogExt for consistency.

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/LocationRecorder.java
dephillipsmichael: Actually, the way I coded this, it requires GPS for accuracy in walking, so I'm going to just get rid of the network provider all together

backbone/src/main/java/org/researchstack/backbone/utils/FormatHelper.java
dephillipsmichael: Believe it or not...

https://www.google.com/search?client=safari&rls=en&q=which+countries+use+the+imperial+unit+system&ie=UTF-8&oe=UTF-8

backbone/src/main/res/layout/rsb_step_layout_active_step.xml
dephillipsmichael: No, I don't think you do.  They wrote all the compat view to magically inject themselves at runtime in place of their legacy counterparts.  However, I've read a few posts that on certain OS' or maybe it was built tools(?) that it didn't work correctly.  Either way, you are right, in xml I don't have to specify this, so I am going to switch it back to ImageView

backbone/src/test/java/org/researchstack/backbone/task/factory/WalkingTaskTests.java
dephillipsmichael: Didn't know about that annotation, thanks for the head's up

PR #351, opened 2017-02-24
Updates for ActivitiesFragment

rianhouston:
Update the UI for Activities to align with iOS.

NOTE:  Before this will work properly, it will need an update to BridgeDataProvider.  However BridgeDataProvider needs some of the model tweaks included here so it is a bit circular.

screenshot_20170224-151402

### DIFF
--- a/skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
+++ b/skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
@@ -1,6 +1,7 @@
 package org.researchstack.skin.ui.fragment;
 
 import android.app.Activity;
+import android.content.Context;
 import android.content.Intent;
 import android.content.res.Resources;
 import android.graphics.drawable.Drawable;
@@ -8,6 +9,7 @@ import android.os.Bundle;
 import android.support.annotation.Nullable;
 import android.support.v4.app.Fragment;
 import android.support.v4.graphics.drawable.DrawableCompat;
+import android.support.v4.widget.SwipeRefreshLayout;
 import android.support.v7.widget.AppCompatTextView;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
@@ -16,8 +18,10 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
+import android.widget.TextView;
 import android.widget.Toast;
 
+import org.joda.time.DateTime;
 import org.researchstack.backbone.StorageAccess;
 import org.researchstack.backbone.result.TaskResult;
 import org.researchstack.backbone.storage.file.StorageAccessListener;
@@ -31,6 +35,7 @@ import org.researchstack.backbone.model.SchedulesAndTasksModel;
 import org.researchstack.skin.ui.views.DividerItemDecoration;
 
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 
@@ -40,10 +45,13 @@ import rx.subjects.PublishSubject;
 
 
 public class ActivitiesFragment extends Fragment implements StorageAccessListener {
+    private static final String LOG_TAG = ActivitiesFragment.class.getCanonicalName();
     private static final int REQUEST_TASK = 1492;
     private TaskAdapter adapter;
     private RecyclerView recyclerView;
     private Subscription subscription;
+    private SwipeRefreshLayout swipeContainer;
+
 
     @Nullable
     @Override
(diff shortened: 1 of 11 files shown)


skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
```diff
@@ -87,30 +110,99 @@ private void setUpAdapter() {
...
-                    });
                 });
     }
 
+    /**
+     * Process the model to create section groups and section headers
+     * @param model
+     * @return
+     */
+    private List<Object> processResults(SchedulesAndTasksModel model) {
```
dephillipsmichael: Make public, in most apps we've done on iOS, we've had to override this method

skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
```diff
@@ -87,30 +110,99 @@ private void setUpAdapter() {
...
+        List<Object> tasks = new ArrayList<>();
+
+        DateTime now = new DateTime();
+        DateTime startOfDay = new DateTime().withTimeAtStartOfDay().minusSeconds(1);
```
dephillipsmichael: What's with the minus 1 seconds?

skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
```diff
@@ -139,76 +231,148 @@ public void onDataAuth() {
...
+            String message;
+
+            public Header(String t, String m) {
+                title = t;
+                message = m;
+            }
+        }
+
+        // TODO: this should live somewhere else like ResourceManager?  Cause it will be app specific
+        //       and the constants defined somewhere?
```
dephillipsmichael: I think it's fine to hard-code certain tasks to colors here for the default behavior, without magic constants, but I think it is important for the user to be able to inject their own ViewHolders easily.  Each app we have done on iOS has had a different Activities UI, so we should focus on making it customizable by the developer.

skin/src/main/res/drawable/rss_ic_circle_empty_24dp.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="oval">
+
+    <size
+        android:width="24dp"
```
dephillipsmichael: We may want to make this a dimension.  An alternate way to do this, that would scale, is to use a ShapeDrawable programattically

skin/src/main/res/layout/rss_fragment_activities.xml
```diff
@@ -1,6 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
-<android.support.v7.widget.RecyclerView xmlns:android="http://schemas.android.com/apk/res/android"
+<android.support.v4.widget.SwipeRefreshLayout
```
dephillipsmichael, line 2: Nice addition

dephillipsmichael:
Looks slick.  My main concern is on customization of the ViewHolders for the headers and normal cells.  If we could push that somehow to a listener and have it overridable in the Fragment, I think that would be a win.

But, as far as blockers go, I don't see any.  Definitely feel free to merge.

skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
rianhouston: Its a bit hacky I agree.  Some of the activites are scheduled on the exact start of the day so the isBefore() and isAfter() methods were not working since it is equal.  This was a quick and dirty solution.

skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
rianhouston: Will do

skin/src/main/res/drawable/rss_ic_circle_empty_24dp.xml
rianhouston: Will do

rianhouston:
I could make it so you can extend or override the entire TaskAdapter.  Do you think that would satisfy your goal for customizable UI?  The view holders and adapter are pretty tightly coupled so changing one may require a change in the other.  I could externalize the TaskAdapter class and allow you to inject your own.

rianhouston:
@dephillipsmichael - I incorporated your feedback.  I don't have permission to merge though so please do so when you can.

skin/src/main/java/org/researchstack/skin/ui/fragment/ActivitiesFragment.java
```diff
@@ -71,14 +79,27 @@ private void unsubscribe() {
...
 
+        fetchData();
+
+    }
+
+    /**
+     * Override this method to provide a customer adapter for your application.
+     * @return  The adapter for displaying the list of tasks.
+     */
+    protected TaskAdapter createTaskAdapter() {
```
dephillipsmichael, line 97: Nice, thanks.  That should be sufficient for someone who want to implement there own UI.

PR #352, opened 2017-02-26
Feature/ma 73

dephillipsmichael:
This PR implements the Tapping Task from the ORKOrderedTask class on iOS.


backbone/src/main/java/org/researchstack/backbone/result/TappingIntervalResult.java
```diff
@@ -0,0 +1,214 @@
...
+    public enum TappingButtonIdentifier {
+        // The touch landed outside of the two buttons.
+        TappingButtonIdentifierNone,
+        // The touch landed in the left button.
+        TappingButtonIdentifierLeft,
+        // The touch landed in the right button.
+        TappingButtonIdentifierRight;
+    }
+
+    public static final class Size implements Serializable {
```
rianhouston: Why did you need to duplicate these classes?  To make them serializable?

backbone/src/main/java/org/researchstack/backbone/task/factory/TappingTaskFactory.java
```diff
@@ -0,0 +1,210 @@
...
+                break;
+            case RIGHT:
+                rightHand = true;
+                break;
+            case BOTH:
+                rightHand = !leftFirstIfDoingBoth;
+                break;
+        }
+
+        // Obtain sensor frequency for Walking Task recorders
```
rianhouston: Walking?  Copy paste?

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/TappingIntervalStepLayout.java
```diff
@@ -0,0 +1,274 @@
...
+                        // Make sure we have the same finger's up tap
+                        if (buttonSamples[idx] != null && correctFingerForIdx) {
+                            buttonSamples[idx].setDuration(motionEvent.getDownTime());
+                            sampleList.add(buttonSamples[idx]);
+                            buttonSamples[idx] = null;
+                            lastPointerIdx[idx] = INVALID_POINTER_IDX;
+                            if (countsAsATap) {
+                                countATap();
+                            }
+                            Log.d("TODO_REMOVE", "tap up with idx " + idx);
```
rianhouston: Use LogExt?

backbone/src/main/java/org/researchstack/backbone/ui/views/FixedSubmitBarLayout.java
```diff
@@ -12,12 +12,15 @@
 
 import org.researchstack.backbone.R;
 import org.researchstack.backbone.ui.step.layout.StepLayout;
+import org.researchstack.backbone.ui.step.layout.TappingIntervalStepLayout;
```
rianhouston: This isn't needed is it?

rianhouston:
Looks good. :shipit:

backbone/src/main/java/org/researchstack/backbone/result/TappingIntervalResult.java
dephillipsmichael: Yeah, they aren't serializable.  You also can't override them and make them serializable because they are final classes.

dephillipsmichael: Maybe some comments for them would be a good idea

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/TappingIntervalStepLayout.java
dephillipsmichael: Whoops, will take this out.

backbone/src/main/java/org/researchstack/backbone/ui/views/FixedSubmitBarLayout.java
dephillipsmichael: Nope, will remove

PR #356, opened 2017-02-28
Feature/ma 74

dephillipsmichael:
Fully implemented the Audio Task as it exists in iOS


backbone/src/main/java/org/researchstack/backbone/step/AudioTooLoudStep.java
```diff
@@ -0,0 +1,95 @@
...
+public class AudioTooLoudStep extends InstructionStep implements NavigableOrderedTask.NavigationSkipRule {
+
+    /**
+     * The step identifier where the AudioResult object will be
+     */
+    private String audioStepResultIdentifier;
+
+    /**
+     * Must be a value from 0.0 - 1.0
+     * Step will be skipped unless the loudness threshold is over this threshold
```
rianhouston, line 31: Is this correct?  Code below seems to indicate step will be skipped if loudness is over the threshold?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/AudioRecorder.java
```diff
@@ -0,0 +1,288 @@
...
+        if (settings == null) {
+            throw new IllegalStateException("Cannot start media recorder since settings is null");
+        }
+
+        // In-app permissions were added in Android 6.0
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            PackageManager pm = context.getPackageManager();
+            int hasPerm = pm.checkPermission(Manifest.permission.RECORD_AUDIO, context.getPackageName());
+            if (hasPerm != PackageManager.PERMISSION_GRANTED) {
+                if (getRecorderListener() != null) {
```
rianhouston, line 91: You fail if permission has not been granted as opposed to asking for permission?

rianhouston:
Looks good to me.  Interested to see the graph in action.

backbone/src/main/java/org/researchstack/backbone/step/AudioTooLoudStep.java
dephillipsmichael: No, this is incorrect, need to reword that comment.

dephillipsmichael: Oh, actually this is correct.  We always skip this step, which says, "noise too loud", unless we want to show it when the loudness is over the threshold.

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/AudioRecorder.java
dephillipsmichael: Trying to manage the permission state from within a Recorder enabled step layout would be asking for trouble.  Some of the active steps start automatically, some of them have a delay, showing anything to the user will disrupt an ActiveStep's user experience.  So, I thought it was a better idea to include a separate PermissionStep before these steps which asks for permission, so it does not interrupt the flow of this step layout.

PR #357, opened 2017-03-05
Feature/ma 75

dephillipsmichael:
Fixed various bugs around packaging and transcribing a TaskResult, and hooked into ActiveStepLayout to upload to the server.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/result/TappingIntervalResult.java
+++ b/backbone/src/main/java/org/researchstack/backbone/result/TappingIntervalResult.java
@@ -1,5 +1,7 @@
 package org.researchstack.backbone.result;
 
+import com.google.gson.annotations.SerializedName;
+
 import java.io.Serializable;
 import java.util.List;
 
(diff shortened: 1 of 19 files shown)


backbone/src/main/java/org/researchstack/backbone/result/TappingIntervalResult.java
```diff
@@ -78,13 +84,15 @@ public void setButtonRect2(Rect buttonRect2) {
...
          */
+        @SerializedName("TapTimeStamp")
         private long timestamp;
 
         /**
          * A duration of the tap event.
          *
          * The duration store time interval between touch down and touch release events.
          */
+        @SerializedName("duration")
```
liujoshua, line 95: Is this really lowercase, while the rest are uppercased?

backbone/src/main/java/org/researchstack/backbone/task/factory/WalkingTaskFactory.java
```diff
@@ -142,7 +143,7 @@ public static OrderedTask shortWalkTask(
                     recorderConfigList.add(new AccelerometerRecorderConfig(AccelerometerRecorderIdentifier, sensorFreq));
                 }
                 if (!optionList.contains(TaskExcludeOption.DEVICE_MOTION)) {
-                    recorderConfigList.add(new AccelerometerRecorderConfig(DeviceMotionRecorderIdentifier, sensorFreq));
+                    recorderConfigList.add(new DeviceMotionRecorderConfig(DeviceMotionRecorderIdentifier, sensorFreq));
```
liujoshua, line 146: Maybe there's a way we could refactor the recorderConfig and exclusion logic to reduce the repeats

liujoshua (approved):
Looks good

PR #358, opened 2017-03-12
Feature/ma 70

dephillipsmichael:
Implemented a TaskItemFactory that can created tasks from JSON similar to how SurveyItem makes Steps from JSON.  Currently, voice, walking, short walk, tremor, and tapping tasks are implemented.

Also in this PR, is implementing a hack to have those tasks be loaded when they are tapped from the ActivitiesFragment.


backbone/src/main/java/org/researchstack/backbone/model/taskitem/factory/TaskItemFactory.java
```diff
@@ -0,0 +1,532 @@
...
+            }
+        }
+        return defaultValue;
+    }
+
+    public Task createCustomTask(Context context, CustomTaskItem item) {
+
+
+
+//        guard let steps = transformTaskSteps(factory) else { return nil }
```
liujoshua, line 440: Are you intending to keep these comments for now? Here and line 460-530

backbone/src/main/java/org/researchstack/backbone/result/TaskResult.java
```diff
@@ -28,22 +28,21 @@
 {
     private Map<String, StepResult> results;
 
-    // unimplemented but exists in RK, implement or delete if not needed
-    private UUID uuidTask;
-
-    // unimplemented but exists in RK, implement or delete if not needed
-    private Uri outputDirectory;
+    /** Store extra details needed for parsing the result in the taskDetails */
+    private Map<String, Object> taskDetails;
```
liujoshua: 👍 thanks!

backbone/src/test/java/org/researchstack/backbone/task/factory/TappingTaskTest.java
```diff
@@ -31,8 +27,8 @@
 
 public class TappingTaskTest {
 
-    private Context mockContext;
-    private Resources mockResources;
+    @Mock private Context mockContext;
```
liujoshua, line 30: To get these annotations to auto-mock, you need to either annotate the Test class with @RunWith(MockitoJUnitRunner.class) or call MockitoAnnotations.initMocks(this); before every test

liujoshua (approved):

backbone/src/test/java/org/researchstack/backbone/task/factory/TappingTaskTest.java
dephillipsmichael: Ohh that makes sense! I thought it was more of a code comment than an injection annotation, but I will do this from now on.

PR #362, opened 2017-03-16
Removed upload code in ActiveTaskActivity. Added logging for debuggin

liujoshua:
Removed upload code from ActiveTaskActivity
In ActivitiesFragment, ActiveTaskActivity launch is changed from startActivity to startActivityForResult,


dephillipsmichael:
approved

PR #370, opened 2017-03-26
Feature/ma 140

dephillipsmichael:
This work was a byproduct of trying to create the TrackedObjectTask in the Bridge SDK.  To create a TrackedObjectTask you need SurveyFactory and TaskItemFactory to be fully customizable.  Up until this point, the ability to customize them has been lightly tested, but as you can see from this PR, it was a major architecture change to get it to work fully for the example.  

This work also introduces a class, TaskCreationManager.  It is meant to be a fusion of the OnboardingManager and TaskProvider, but updated to support fully customization of Tasks through JSON.


backbone/src/main/java/org/researchstack/backbone/ResourceManager.java
```diff
@@ -113,4 +122,53 @@ public static ResourceManager getInstance() {
...
+    public static class NameJsonProvider implements ResourceNameToStringConverter {
+
+        private Context context;
+
+        public NameJsonProvider(Context context) {
+            this.context = context;
+        }
+
+        @Override
+        public String getJsonStringForResourceName(String resourceName) {
```
liujoshua: Looks like getJsonStringForResourceName and getHtmlStringForResourceName both handle any sort of string resource, not just JSON. 

Does this exist basically to work around needing to implement and instantiate ResourcePathManager.Resource? IMO it'd be worth simplifying ResourcePathManager, the implementation of typing is incomplete and feels cludgy (e.g. HTML vs HTM file extension, Resource.types are static variables, no extension/inheritance) and don't seem to add much value. 

Thoughts?

backbone/src/main/java/org/researchstack/backbone/task/TaskCreationManager.java
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * Created by TheMDP on 3/24/17.
+ *
+ * The TaskCreationManager follows a similar architecture as the OnboardingManager,
+ * Its job is to parse JSON resources and create Tasks from them
+ * It also is designed in a way that makes creating custom tasks and steps easy
+ * while still getting all the awesome features of the base Survey and TaskItem factories
+ */
+
+public class TaskCreationManager implements TaskItemFactory.CustomTaskCreator, SurveyFactory.CustomStepCreator {
```
liujoshua, line 28: nice

liujoshua (approved):
Looks good, one comment/suggestion about Resources (which I know you didn't really touch in this PR)

backbone/src/main/java/org/researchstack/backbone/ResourceManager.java
dephillipsmichael: I'm glad you called this out and I was able to take a second look at it.  It does seem like I was trying to make the ResourceManager more open-ended so that all you needed was a filename and you could get a path to a file, instead of having to add an abstract method every time one is needed.  It seems like a better approach is to refactor ResourceManager so that it can be more extensible.  I'll work on that before merging in this PR.

liujoshua: Cool. My though on this is the Resource.type thing doesn't help us much. We could access Resources the standard way, providing the full filename + extension, and have helpers such as getResourceAsString getResourceAsBytes, getResourceAsJsonObject, getResourceAsCertificate

dephillipsmichael: I agree with you, I'll make those changes.

dephillipsmichael: Changes made, let me know what you think @liujoshua

PR #373, opened 2017-04-05
bug/path

dephillipsmichael:
Fixed bug in path reference, generateAbsolutePath is only meant for WebViews


dephillipsmichael:
I'm going to merge this in as it's a breaking fix

PR #388, opened 2017-10-20
public_class

dephillipsmichael:
Made JsonArrayDataRecorder public for other developers to use along with other public accessors.

I also encapsulated some existing functionality in ActiveStepLayout to allow for custom recorders.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ActiveStepLayout.java
+++ b/backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ActiveStepLayout.java
@@ -193,19 +193,31 @@ public class ActiveStepLayout extends FixedSubmitBarLayout
         }
 
         recorderList = new ArrayList<>();
-        File outputDir = getContext().getFilesDir();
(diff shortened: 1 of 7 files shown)


backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ActiveStepLayout.java
```diff
@@ -189,22 +193,44 @@ protected void start() {
...
 
         recorderList = new ArrayList<>();
+        File outputDir = getOutputDirectory();
+
+        if (activeStep.getRecorderConfigurationList() != null) {
+            for (RecorderConfig config : getRecorderConfigurationList()) {
+                Recorder recorder = config.recorderForStep(activeStep, outputDir);
+                // recorder can be null if it requires custom setup
+                if (recorder == null) {
+                    recorder = createCustomRecorder(config);
```
dephillipsmichael, line 203: I added custom recorder setup functionality using an overridable class.  This is needed for the heart rate camera recorder, since it passes in a SurfaceView for the camera to attach to that is required for the recorder.

PR #392, opened 2017-10-25
Inject ui

dephillipsmichael:
I made some changes to the structure of fixed submit bar layout and instruction step layout so that you can plug in a unique xml layout but keep their functionality.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/ui/ViewTaskActivity.java
+++ b/backbone/src/main/java/org/researchstack/backbone/ui/ViewTaskActivity.java
@@ -54,13 +54,13 @@ public class ViewTaskActivity extends PinCodeActivity implements StepCallbacks
     {
         super.onCreate(savedInstanceState);
         super.setResult(RESULT_CANCELED);
-        super.setContentView(R.layout.rsb_activity_step_switcher);
+        super.setContentView(getContentViewId());
 
         toolbar = (Toolbar) findViewById(R.id.toolbar);
         setSupportActionBar(toolbar);
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
 
-        root = (StepSwitcher) findViewById(R.id.container);
+        root = (StepSwitcher) findViewById(getViewSwitcherRootId());
 
         if(savedInstanceState == null)
         {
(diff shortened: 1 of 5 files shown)


backbone/src/main/java/org/researchstack/backbone/ui/ViewTaskActivity.java
```diff
@@ -91,6 +91,14 @@ public Step getCurrentStep()
         return currentStep;
     }
 
+    public int getContentViewId() {
```
liujoshua: Nice to have: `@LayoutRes` annotation

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/InstructionStepLayout.java
```diff
@@ -97,19 +102,20 @@ public void onDetachedFromWindow() {
         }
     }
 
-    private void initializeStep() {
-
-        titleTextView       = (TextView)findViewById(R.id.rsb_intruction_title);
-        textTextView        = (TextView)findViewById(R.id.rsb_intruction_text);
-        imageView           = (ImageView) findViewById(R.id.rsb_image_view);
-        moreDetailTextView  = (TextView)findViewById(R.id.rsb_instruction_more_detail_text);
+    public void connectStepUi(int titleRId, int textRId, int imageRId, int detailRId) {
```
liujoshua: Nice to have: Annotation with `@IdRes`

liujoshua (approved):
Looks good. Not sure of the use cases, will want to see them. (Building similar things on RSuite)

Is the idea that an app may have multiple extends of ViewTaskActivity that override the layout in different ways? If each app just needs to customize this view, might be easier to just override rsb_activity_step_switcher

backbone/src/main/java/org/researchstack/backbone/ui/ViewTaskActivity.java
dephillipsmichael: Yes thanks!  I was trying to remember the annotation for that and kept coming up with nothing.

dephillipsmichael:
It's not really the goal to have multiple extends of ViewTaskActivity, but to allow for greater customization of the UI, while still taking advantage of all the functionality code within ViewTaskActivity.  Also, I should add a way to just override ViewSwitcher, which I will add now, but in the case of CRF, we need to remove and alter the Toolbar as well.

PR #402, opened 2017-11-05
feature/stepResultHelper

dephillipsmichael:
Added findStringResult method


DwayneJengSage (commented):
Should we add unit tests to StepResultHelperTests? What are the requirements on testing null checks for args? Should we have a test case for both the found case and not found case?

dephillipsmichael:
Unit tests already exist for the methods in this class see StepResultHelperTest. The method I added just piggy backs off another method, I could add a unit test but don't really see the value

DwayneJengSage (approved):
As long as it's been tested one way or another, it looks fine to me.

PR #403, opened 2017-11-10
login with external ID

DwayneJengSage:
Adding page for login with external ID.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItemType.java
+++ b/backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItemType.java
@@ -57,8 +57,6 @@ public enum SurveyItemType {
     ACCOUNT_EMAIL_VERIFICATION  ("emailVerification"   ),   // EmailVerificationStep
     @SerializedName("externalID")
     ACCOUNT_EXTERNAL_ID         ("externalID"          ),   // ExternalIDStep
-    @SerializedName("externalIDLogin")
-    ACCOUNT_EXTERNAL_ID_LOGIN   ("externalIDLogin"     ),   // ExternalIDLoginStep
     @SerializedName("permissions")
     ACCOUNT_PERMISSIONS         ("permissions"         ),   // PermissionsStep
     @SerializedName("onboardingCompletion")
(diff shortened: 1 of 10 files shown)


backbone/src/main/java/org/researchstack/backbone/model/survey/SurveyItemType.java
```diff
@@ -57,6 +57,8 @@
     ACCOUNT_EMAIL_VERIFICATION  ("emailVerification"   ),   // EmailVerificationStep
     @SerializedName("externalID")
     ACCOUNT_EXTERNAL_ID         ("externalID"          ),   // ExternalIDStep
```
liujoshua: ACCOUNT_EXTERNAL_ID, which currently produces a non-implemented step, is intended to support this feature. We should use this. 

@dephillipsmichael, does that sound right? 

See:

https://github.com/Sage-Bionetworks/BridgeAppSDK/blob/6a135f98a2e7c4020f278a76fcdf97ff2f5fc777/BridgeAppSDKSample/StudyOverviewViewController.swift#L63

backbone/src/main/java/org/researchstack/backbone/model/survey/factory/SurveyFactory.java
```diff
@@ -690,8 +724,9 @@ public ProfileStep createProfileStep(Context context, ProfileSurveyItem item) {
      * @param item InstructionSurveyItem from JSON
```
liujoshua, line 722: Please add @param loginOptions

backbone/src/main/java/org/researchstack/backbone/model/survey/factory/SurveyFactory.java
```diff
@@ -84,6 +86,16 @@
...
     public static final String CONSENT_QUIZ_IDENTIFIER = "consentQuiz";
 
+    @VisibleForTesting
+    static final int EXTERNAL_ID_MAX_LENGTH = 128;
+
+    private static final List<ProfileInfoOption> EXTERNAL_ID_LOGIN_OPTIONS;
+    static {
+        List<ProfileInfoOption> tempList = new ArrayList<>();
+        tempList.add(ProfileInfoOption.EXTERNAL_ID);
+        EXTERNAL_ID_LOGIN_OPTIONS = Collections.unmodifiableList(tempList);
```
liujoshua, line 96: +1 for immutable

liujoshua (changes requested):
Thanks, looks great. One change request for consistency with ResearchKit constants

liujoshua (approved):

PR #419, opened 2017-11-23
Broadcast DataProvider errors, and handle them when BaseActivities are visible

liujoshua:


skin/src/main/java/org/researchstack/skin/AppPrefs.java
```diff
@@ -52,22 +56,37 @@ public static AppPrefs getInstance()
...
         boolean isAutoLocked = prefs.getBoolean(SettingsFragment.KEY_AUTO_LOCK_ENABLED, true);
 
         String time = prefs.getString(SettingsFragment.KEY_AUTO_LOCK_TIME, "1");
         int autoLockMins = isAutoLocked ? Integer.parseInt(time) : 60 * 24 * 365;
         return autoLockMins * 60 * 1000;
     }
 
-    public void setSkippedOnboarding(boolean skipped)
-    {
+    public void addDataProviderError(@Nullable @DataProvider.Errors String dataProviderError) {
```
dephillipsmichael, line 67: Perhaps some explanation here in comments or method header on why a developer would want to save error strings?

skin/src/main/java/org/researchstack/skin/DataProviderErrorReceiver.java
```diff
@@ -0,0 +1,128 @@
...
+
+        if (AppPrefs.getInstance().skippedOnboarding()) {
+            // We don't want to bother a user that has skipped sign-up with the signed out
+            // or consent messages. Short-circuiting until we have an approved message to show
+            // a user that has skipped.
+            return;
+        }
+
+        if (!(context instanceof BaseActivity)) {
+            LogExt.w(getClass(), "Received by class that is not a BaseActivity");
```
dephillipsmichael, line 46: The only Activity that we really use that extends BaseActivity is MainActivity, and even that is only used to show activities and dashboard data.  Maybe consider instanceof PinCodeActivity to have a broader reach?

skin/src/main/java/org/researchstack/skin/DataProviderErrorReceiver.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    public static void handleError(@NonNull Context context,
+                                   @NonNull @DataProvider.Errors String error) {
+        checkNotNull(context);
+        checkNotNull(error);
+
+        if (!Strings.isNullOrEmpty(error)) {
+            AppPrefs.getInstance().addDataProviderError(error);
```
dephillipsmichael, line 66: Did you think about sending the error data as intent data?

skin/src/main/java/org/researchstack/skin/DataProviderErrorReceiver.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        LocalBroadcastManager.getInstance(context)
+                .sendBroadcast(new Intent(ACTION_DATA_PROVIDER_ERROR));
+    }
+
+
+    @VisibleForTesting
+    void handleInActivity(BaseActivity activity) {
```
dephillipsmichael, line 74: I'm not seeing anywhere in this method that you specifically need "BaseActivity" vs any "Activity".  Is there a reason for this?

skin/src/main/java/org/researchstack/skin/DataProviderErrorReceiver.java
liujoshua: There maybe not be an activity available to handle the broadcast, so we want to persist the errors. 

I didn't see much advantage to reading in errors and sending them in the message versus using the message as a trigger and reading in the errors.

Also, this edge case seems unlikely, but if unhandled errors built up, they should be handled with different priorities, e.g. re-authentication should happen before re-consent. So we want to ensure we're looking at the set of unhandled errors while the order of the broadcast messages/errors can be totally ignored.

skin/src/main/java/org/researchstack/skin/DataProviderErrorReceiver.java
liujoshua: Good point, PinCodeActivity is the right place for this

skin/src/main/java/org/researchstack/skin/DataProviderErrorReceiver.java
liujoshua: It's currently using methods/views in BaseActivity, but I think it could work on any activity

liujoshua:
Waiting on refactor of BaseActivity and PinCodeActivity

PR #424, opened 2017-12-02
bug/AA-8

dephillipsmichael:
Fixed consent sharing StepResult parsing and use in ConsentReview


backbone/src/main/java/org/researchstack/backbone/ui/ViewTaskActivity.java
```diff
@@ -369,4 +373,8 @@ public void setActionBarTitle(String title)
             actionBar.setTitle(title);
         }
     }
+
+    public interface ResultListener {
```
liujoshua, line 377: Nice. This will be a place to hook in that JSON serialization for saving clientData with ScheduleActivity

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ConsentReviewSubstepListStepLayout.java
```diff
@@ -99,8 +114,29 @@ protected ConsentSignatureBody createConsentSignatureBody(StepResult stepResult)
         }
         body.birthdate = usersBirthday;
 
+        // If one doesn't exist yet then set it to a blank string, otherwise the consent sig will have null value issues
```
liujoshua, line 117: So, if sharingScope is null, but body.scope is not null, we want to clear out the body.scope? 

Does this work because setting the sharing scope is currently the last step? Do we have to clear out body.scope to have correct behavior? Seems like maybe if you had partially gone through consent, body.scope could be used to pre-populate the sharing step or something. Not required, just trying to think through what this does

dephillipsmichael: Nice catch, this is a bug, it should be....

if (sharingScope == null) {
        if (body.scope == null) {
             sharingScope = "";
        }
} else {
   body.scope = sharingScope;
}

backbone/src/main/java/org/researchstack/backbone/ui/ViewTaskActivity.java
dephillipsmichael: Yeah I think so too.  Just make sure the "instanceof" check and the call to that listener is in a spot where it is called on the previous StepLayout (so it's result is set)

PR #427, opened 2017-12-04
Refactor skin into backbone

liujoshua:


dephillipsmichael:
@liujoshua there is a conflict but otherwise this should be good to merge

PR #430, opened 2017-12-10
fix title display bug, fix intruction typos

EricSiegNW:
Mainly pushing to get rid of the bug created by the "if no title, they must have screwed up" code. Sometimes we just don't want a title. If we wanted a title we'd have something in the title field.

Also addressing some intruction_title etc typos while I was there since it was bothering me.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/ui/step/layout/InstructionStepLayout.java
+++ b/backbone/src/main/java/org/researchstack/backbone/ui/step/layout/InstructionStepLayout.java
@@ -4,8 +4,6 @@ import android.annotation.TargetApi;
 import android.content.Context;
 import android.content.Intent;
 import android.content.res.Resources;
-import android.graphics.drawable.Animatable;
-import android.graphics.drawable.Animatable2;
 import android.graphics.drawable.AnimationDrawable;
 import android.graphics.drawable.Drawable;
 import android.os.Handler;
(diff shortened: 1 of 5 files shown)


backbone/src/main/java/org/researchstack/backbone/ui/step/layout/InstructionStepLayout.java
```diff
@@ -127,14 +124,6 @@ public void refreshStep() {
             String title = step.getTitle();
             String text  = step.getText();
 
-            if (TextUtils.isEmpty(title) &&
-                    !TextUtils.isEmpty(text) && !TextUtils.isEmpty(instructionStep.getMoreDetailText()))
-            {
-                // With no Title, we can assume text and detail text is equla to title and text
```
EricSiegNW, line 133: This is the main point of the change. Took a bit to figure out why things were acting broken when I didn't supply a title, even though I didn't want a title.

backbone/src/test/java/org/researchstack/backbone/step/SubtaskStepTests.java
```diff
@@ -75,14 +75,14 @@ public void testMutatedResultSet() throws Exception
 
         Step step1 = navTask.getStepAfterStep(null, taskResult);
         assertNotNull(step1);
-        assertEquals(step1.getIdentifier(), "intruction");
+        assertEquals(step1.getIdentifier(), "instruction");
```
dephillipsmichael, line 78: I miss your grammar police coding! Welcome back to Android :)

backbone/src/main/java/org/researchstack/backbone/ui/step/layout/InstructionStepLayout.java
dephillipsmichael: I'm glad you had the initiative to change this!  I found this during CRF dev and was lazy and just made title be a blank String to "fix".  I was afraid this would break things in other apps, but this really shouldn't be there, it's super confusing.

PR #449, opened 2018-01-08
Feature/archive result format

liujoshua:


backbone/src/main/java/org/researchstack/backbone/step/active/recorder/DeviceMotionRecorder.java
```diff
@@ -70,7 +73,16 @@ public void start(Context context) {
     @Override
     protected void writeJsonData() {
         // Update the main json object
-        jsonObject.addProperty(TIMESTAMP_KEY, System.currentTimeMillis());
+        jsonObject.addProperty(TIMESTAMP_IN_SECONDS_KEY, System.currentTimeMillis() * 1e-3d);
+
+        double uptime;
```
dephillipsmichael: I remember from my meeting with Shannon that "uptime" was in fact elapsedRealtime from cpu boot time, but I thought we didn't need that because we had System.currentTimeMillis().  However, now that I think about it more, grabbing the system time at the point of writeJsonData() is a bug, and not accurate of when the sensor actually recorded the value.

Also, on further investigation, Android also uses "uptime" the same way iOS does for it's sensors.  See this post for more information https://stackoverflow.com/questions/5500765/accelerometer-sensorevent-timestamp

What we need to change this to is remove any timestamp setting in "writeJsonData" and move it to each recorder's onSensorChanged(SensorEvent sensorEvent), and grab the "uptime" value and set it there. 

Does this make sense?

liujoshua: Yup, all makes sense, thanks.

liujoshua: Seems then we should write the JSON on the sensor event and not use writeJsonData() and startRunnableForWritingJson() to control frequency.

PedometerRecorder uses the "manual frequency", but it's implementation is writing on onSensorEvent also, so that shouldn't cause an issue.

dephillipsmichael: We can do that for individual sensors just fine.  However, my goal was to group them together like in DeviceMotionSensor, and I didn't want to write all the data each time a single sensor changed it's value.  Does that make sense?  Any ideas on how to do that?

liujoshua:
@dephillipsmichael I've made some significant changes. Could you please take another look?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/DeviceMotionRecorder.java
```diff
@@ -101,50 +119,167 @@ protected void writeJsonData() {
...
+        String sensorTypeKey = SENSOR_TYPE_TO_DATA_TYPE.get(sensorType);
+
+        if (Strings.isNullOrEmpty(sensorTypeKey)) {
+            logger.warn("Unable find type key for sensor type: "
+                    + sensorType);
+            return;
+        }
+
+        jsonObject.addProperty(SENSOR_DATA_TYPE_KEY, sensorTypeKey);
+        jsonObject.addProperty(SENSOR_EVENT_ACCURACY_KEY, sensorEvent.accuracy);
```
dephillipsmichael, line 195: Shouldn't you add "uptime" here?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/PedometerRecorder.java
```diff
@@ -128,16 +123,18 @@ public void onStepTaken() {
...
         // Step detector is only available for OS kitkat and above
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT &&
-            sensorEvent.sensor.getType() == Sensor.TYPE_STEP_DETECTOR)
+                sensorEvent.sensor.getType() == Sensor.TYPE_STEP_DETECTOR)
         {
             onStepTaken();
         } else if (sensorEvent.sensor.getType() == Sensor.TYPE_ACCELEROMETER) {
             accelerometerStepDetector.processAccelerometerData(sensorEvent);
         }
+        //TODO: fix accelerometer data
```
dephillipsmichael, line 137: What's wrong with accel data?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/SensorRecorder.java
```diff
@@ -86,54 +87,66 @@ public void start(Context context) {
...
+            // set timestamp reference, which timestamps are measured relative to
+            timestampZeroReferenceNanos = sensorEvent.timestamp;
+
+            // record date equivalent of timestamp reference
+            long uptimeNanos;
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+                uptimeNanos = SystemClock.elapsedRealtimeNanos();
+            } else {
+                uptimeNanos = SystemClock.elapsedRealtime() * 100000; // millis to nanos
```
dephillipsmichael: / 1e-6

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/SensorRecorder.java
```diff
@@ -86,54 +87,66 @@ public void start(Context context) {
...
+            }
+
+            long timestampReferenceMillis = System.currentTimeMillis()
+                    + ((timestampZeroReferenceNanos - uptimeNanos) / 1000000L);
```
dephillipsmichael: This value 1000000L has 6 zeros, while the value above has 5 zeros.  Intentional?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/SensorRecorder.java
```diff
@@ -86,54 +87,66 @@ public void start(Context context) {
...
+            Date timestampReferenceDate = new Date(timestampReferenceMillis);
+            jsonObject.addProperty(TIMESTAMP_DATE_KEY,
+                    new SimpleDateFormat(FormatHelper.DATE_FORMAT_ISO_8601, Locale.getDefault())
+                            .format(timestampReferenceDate));
         }
+
+        jsonObject.addProperty(TIMESTAMP_IN_SECONDS_KEY,
+                (sensorEvent.timestamp - timestampZeroReferenceNanos) * 1e-9);
```
dephillipsmichael: long * double = double, correct?

dephillipsmichael:
This looks awesome, nice work Josh.

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/DeviceMotionRecorder.java
liujoshua: I put that in SensorRecorder, so everything that extends it will use the same timestamp and uptime.

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/PedometerRecorder.java
liujoshua: This was a reminder to change AccelerometerRecorder to use the new format of DeviceMotion

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/SensorRecorder.java
liujoshua: Thanks

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/SensorRecorder.java
liujoshua: Thanks

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/SensorRecorder.java
liujoshua: Yes, added a comment

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/AccelerometerRecorder.java
```diff
@@ -38,39 +24,6 @@
         if (hasAvailableType(availableSensorList, Sensor.TYPE_ACCELEROMETER)) {
             return Collections.singletonList(Sensor.TYPE_ACCELEROMETER);
         }
-        return new ArrayList<>();
```
dephillipsmichael, line 41: Nice idea!  You got this class down to like 2 lines.

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/JsonArrayDataRecorder.java
```diff
@@ -67,7 +67,7 @@ public void onWriteComplete(File file) {
         startTime = System.currentTimeMillis();
 
         // Since we are writing a JsonArray, have the header and footer be
-        dataLogger.start("{\"" + JSON_ITEMS_KEY + "\":[", "]}");
+        dataLogger.start("[", "]");
```
dephillipsmichael, line 70: iOS doesn't do the items key anymore?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/PedometerRecorder.java
dephillipsmichael: I still see this line, is this fixed?

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/SensorRecorder.java
```diff
@@ -86,54 +86,67 @@ public void start(Context context) {
...
-                        // minimal write delay to give the UI thread some time to catch up
-                        // and hopefully get the frequency back up to the desired one
-                        delayGoal = 1;
-                    }
-
-                    mainHandler.postDelayed(jsonWriterRunnable, delayGoal);
-                }
-            };
-            mainHandler.postDelayed(jsonWriterRunnable, writeDelayGoal);
+        JsonObject jsonObject = new JsonObject();
```
dephillipsmichael, line 106: This is a solid architecture improvement.  Nice work!

PR #450, opened 2018-01-10
feature/AA-90

dephillipsmichael:
Added spoken instruction map to match iOS JSON functionality


backbone/src/main/java/org/researchstack/backbone/ui/step/layout/ActiveStepLayout.java
```diff
@@ -313,29 +316,63 @@ protected void startAnimation() {
...
+            String countdownKey = "countdown";
+            if (activeStep.getSpokenInstructionMap().containsKey(countdownKey)) {
+                String countdownValStr = activeStep.getSpokenInstructionMap().get(countdownKey);
+                try {
+                    int countdownTime = Integer.parseInt(countdownValStr);
+                    if (secondsLeft <= countdownTime && secondsLeft > 0) {
+                        speakText(String.valueOf(secondsLeft));
+                    }
+                } catch (NumberFormatException e) {
+                    e.printStackTrace();
```
liujoshua: LogExt instead of printStackTrace

liujoshua (approved):
Looks good, one small change

PR #451, opened 2018-01-13
Feature/AA-140

dephillipsmichael:
This PR solves adds support for guaranteed recording when the screen is turned off, or when the activity is destroyed/re-created for whatever reason that the system decided to.

It is important when testing this branch, to debug using this developer option in the Android system settings called "Do not keep activities" that will simulate the scenario of the system destroying your activity in the background like when the screen is off for a long running active step.  Tasks should be tested with that debugging option, and without it.

To accomplish background recording, I moved all recording and TTS functionality to a service called RecorderService.  The logic of this service is very similar to what ActiveStepLayout was doing previously.  Except now, when we needed to call methods in ActiveStepLayout, we send Broadcasts to communicate.  Also, all recorders now have been switched to using the Broadcasts to communicate state updates.

Most of the ActiveStepLayout sub-classes didn't need any changes except for switching to broadcast paradigm.  I tested them all and they work as they did before expected.

The one sub-class that needed significant logic changes was CameraActiveStepLayout.  It enforces that the ActiveStepLayout be destroyed and the flow moved to the previous step, if the screen is turned off, or activity isPaused for any reason.  The Recorder does not get sent to the RecorderService to achieve this, but is simply managed by the ActiveStepLayout like it used to be.

Overall, this is a huge improvement, and from my testing, it works flawlessly with the screen off, in the pocket of the user, when they do long walking tasks.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/step/active/RecorderService.java
+++ b/backbone/src/main/java/org/researchstack/backbone/step/active/RecorderService.java
@@ -33,7 +33,6 @@ import android.os.Handler;
 import android.os.IBinder;
 import android.os.Vibrator;
 import android.speech.tts.TextToSpeech;
-import android.support.annotation.DrawableRes;
 import android.support.annotation.Nullable;
 import android.support.annotation.RequiresPermission;
 import android.support.v4.app.NotificationCompat;
@@ -50,7 +49,6 @@ import org.researchstack.backbone.step.active.recorder.Recorder;
 import org.researchstack.backbone.step.active.recorder.RecorderConfig;
 import org.researchstack.backbone.step.active.recorder.RecorderListener;
 import org.researchstack.backbone.task.Task;
-import org.researchstack.backbone.ui.ActiveTaskActivity;
 import org.researchstack.backbone.ui.ViewTaskActivity;
 import org.researchstack.backbone.utils.LogExt;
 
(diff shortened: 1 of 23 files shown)


backbone/src/main/java/org/researchstack/backbone/step/active/RecorderService.java
```diff
@@ -0,0 +1,648 @@
...
+        }
+        // This was a valid recorder result that was just read, so consume it
+        prefs.edit().remove(key).apply();
+        return resultHolder;
+    }
+
+    /**
+     * @param appContext
+     * @param activeStep to use to search for a previous start time
+     * @return null if the active step recording has not been started, null otherwise
```
liujoshua, line 122: Change "null otherwise"

backbone/src/main/java/org/researchstack/backbone/step/active/RecorderService.java
```diff
@@ -0,0 +1,648 @@
...
+    private void startDelayedOperations() {
+        // Now allow the recorder to record for as long as the active step requires
+        mainHandler.postDelayed(this::onRecorderDurationFinished,
+                activeStep.getStepDuration() * 1000L);
+
+        startSpeechToTextMap();
+    }
+
+    protected void startSpeechToTextMap() {
+        String endKey = "end";
```
liujoshua: Let's make these static constants

backbone/src/main/java/org/researchstack/backbone/step/active/recorder/RecorderListener.java
```diff
@@ -30,4 +33,9 @@
      * @param error           The error that occurred.
      */
     void onFail(Recorder recorder, Throwable error);
+
+    /**
+     * @return a valid Context for the recorder to broadcast status, null if not available
+     */
+    @Nullable Context onBroadcastContextRequested();
```
liujoshua: getBroadcastContext()

liujoshua (approved):
Great work. Thanks for all the improvements

PR #452, opened 2018-01-23
Add survey item for email-only sign up and sign in

liujoshua:


backbone/src/main/java/org/researchstack/backbone/DataProvider.java
```diff
@@ -78,13 +80,25 @@ public static void init(DataProvider instance) {
...
+            password);
+
+    /**
+     * Request for a link to sign in user
+     *
+     * @param username the user's username
+     * @return Observable of the result of the method, with {@link DataResponse#isSuccess()}
+     * returning true if signIn was successful
+     */
+    public abstract Observable<DataResponse> requestSignInLink(String username);
```
dephillipsmichael: Having to maintain an abstract DataProvider is difficult.  Does it make sense to enforce the requestSignInLink(String username) for all devs?

liujoshua:
Reopening for Travis

PR #461, opened 2018-02-01
Feature/email only auth

dephillipsmichael:
This PR adds email-only sign up and sign in functionality to ResearchStask.  I have tested this through BP app with email-only and normal email/password and both work as expected.

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/utils/SerializableUtils.java
+++ b/backbone/src/main/java/org/researchstack/backbone/utils/SerializableUtils.java
@@ -30,7 +30,7 @@ public class SerializableUtils {
             objectOutputStream.close();
             encoded = Base64.encodeToString(byteArrayOutputStream.toByteArray(), 0);
         } catch (IOException e) {
-            e.printStackTrace();
+            LogExt.e(SerializableUtils.class, e.getLocalizedMessage());
         }
         return encoded;
     }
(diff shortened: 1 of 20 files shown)


backbone/src/main/java/org/researchstack/backbone/utils/SerializableUtils.java
```diff
@@ -0,0 +1,57 @@
...
+    public static String toBase64String(Serializable object) {
+        String encoded = null;
+        try {
+            ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+            ObjectOutputStream objectOutputStream = new ObjectOutputStream(byteArrayOutputStream);
+            objectOutputStream.writeObject(object);
+            objectOutputStream.close();
+            encoded = Base64.encodeToString(byteArrayOutputStream.toByteArray(), 0);
+        } catch (IOException e) {
+            e.printStackTrace();
```
liujoshua: Should these go to logcat?

backbone/src/main/java/org/researchstack/backbone/utils/SerializableUtils.java
```diff
@@ -0,0 +1,57 @@
...
+     * @return the serializable object, or null if something went wrong
+     */
+    public static Serializable fromBase64String(String base64String) {
+        byte[] bytes = Base64.decode(base64String,0);
+        Serializable object = null;
+        try {
+            ObjectInputStream objectInputStream = new ObjectInputStream( new ByteArrayInputStream(bytes) );
+            object = (Serializable)objectInputStream.readObject();
+        } catch (IOException e) {
+            e.printStackTrace();
```
liujoshua: Log for these also

liujoshua (approved):
Minor error logging comments

backbone/src/main/java/org/researchstack/backbone/utils/SerializableUtils.java
dephillipsmichael: Yes good idea, was copy/paste code

PR #465, opened 2018-02-13
Bug/AABP-297

dephillipsmichael:
This method was being called twice in an edge case when the user would tap back button very fast to leave a task.  Do a null check to avoid the crash.


liujoshua (approved):
We should probably disable the button too

dephillipsmichael:
I thought about that, but if you disabled the back button you would also have the side effect of not allowing a step to move backwards more than one step at a time.

PR #467, opened 2018-02-19
Feature/clean up

liujoshua:


backbone/src/main/java/org/researchstack/backbone/model/SchedulesAndTasksModel.java
```diff
@@ -15,6 +15,7 @@
         public String delay;
         public String scheduleString;
         public Date scheduledOn;
+        public Date expiresOn;
```
dephillipsmichael, line 18: It seems at this point it would be valuable to focus on an easy way for SchedulesAndTasksModel to be sub-classed so we could have BridgeSchedulesAndTasksModel.  I'm sure other devs will need to add fields as well, but don't have access to the source.

dephillipsmichael:
Looks good to me, but CI failed

backbone/src/main/java/org/researchstack/backbone/model/SchedulesAndTasksModel.java
liujoshua: I agree that'd be good. ResearchSuite aims to clean things up and make models easily extensible.

PR #472, opened 2018-03-23
Don't allow DataLogger writes after stop or cancelled

liujoshua:
Guard against behavior seen on slower devices


dephillipsmichael:
Looks good to me!

PR #476, opened 2018-04-25
Request phone sign

EricSiegNW:
Add request phone sign in method

### DIFF
--- a/backbone/src/main/java/org/researchstack/backbone/DataProvider.java
+++ b/backbone/src/main/java/org/researchstack/backbone/DataProvider.java
@@ -101,6 +101,19 @@ public abstract class DataProvider {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


backbone/src/main/java/org/researchstack/backbone/DataProvider.java
```diff
@@ -100,6 +100,19 @@ public static void init(DataProvider instance) {
         return Observable.just(new DataResponse(false, "Not implemented"));
     }
 
+    /**
+     * Request for an SMS for the user to login or sign up
+     *
+     * @param regionCode the user's region code
```
dephillipsmichael, line 106: It may be helpful to add more comments for region code like "i.e. XXX or XXX"

dephillipsmichael:
approved

PR #478, opened 2018-06-18
update app compat, notifications, support library

EricSiegNW:


dephillipsmichael:
I believe you also have to update the travis CI files when you do anything with libraries.

PR #481, opened 2018-10-05
Feature/studio 3 2

liujoshua:


liujoshua:
Verified connectedCheck succeeds locally (something we haven't been doing)

PR #484, opened 2018-10-15
Added abstract function to get license file

aishwarya-rm:
Added an abstract function so that MpResourceManager (which extends ResourceManager) can access the License html file and display it in the Profile page.


dephillipsmichael:
Good idea, every app should have a license.

dephillipsmichael (approved):

PR #485, opened 2018-10-16
Support for additional TaskModel Survey element fields

dephillipsmichael:


backbone/src/main/java/org/researchstack/backbone/model/survey/ChoiceQuestionSurveyItem.java
```diff
@@ -8,7 +8,7 @@
 
 public class ChoiceQuestionSurveyItem extends QuestionSurveyItem<Choice> {
     /* Default constructor needed for serilization/deserialization of object */
-    ChoiceQuestionSurveyItem() {
+    public ChoiceQuestionSurveyItem() {
```
liujoshua, line 11: Do we need to increase visibility of these constructors?

backbone/src/main/java/org/researchstack/backbone/ui/step/body/TextQuestionBody.java
```diff
@@ -27,13 +27,13 @@
     //-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
     // Constructor Fields
     //-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*
-    private QuestionStep step;
-    private StepResult<String> result;
+    protected QuestionStep step;
```
liujoshua, line 30: Do we need to increase visibility? Should we make the step and result fields final?

liujoshua (approved):
LGTM, just some questions around visibility

backbone/src/main/java/org/researchstack/backbone/model/survey/ChoiceQuestionSurveyItem.java
dephillipsmichael: Yes, otherwise there are no constructors for it (only GSON)

backbone/src/main/java/org/researchstack/backbone/ui/step/body/TextQuestionBody.java
dephillipsmichael: Yes, sub-classes should have access to these.

PR #488, opened 2018-11-13
Feature/project config

jdkizer9:
Project config updates
Added module for Foundation layer


skin/build.gradle
```diff
@@ -173,3 +191,6 @@ bintray {
         publish = true
     }
 }
+repositories {
```
liujoshua: I think we could put this in the top level `build.gradle`'s allprojects repositories block

build.gradle
```diff
@@ -1,19 +1,26 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext {
+        kotlin_version = '1.2.71'
```
liujoshua: I've found it nice to also have a variable for `support_library_version`

build.gradle
```diff
@@ -1,19 +1,26 @@
...
     dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         // change this to 1.5.x if you're not on Android Studio 2.0.0 beta
-        classpath 'com.android.tools.build:gradle:2.1.3'
-        classpath 'com.github.dcendents:android-maven-gradle-plugin:1.4.1'
-        classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.6'
-        classpath 'me.tatarka:gradle-retrolambda:3.2.3'
-        classpath "com.neenbedankt.gradle.plugins:android-apt:1.4"
+        classpath 'com.android.tools.build:gradle:3.2.1'
+        classpath 'me.tatarka:gradle-retrolambda:3.7.0'
```
liujoshua: Looks like we can remove this now :)

build.gradle
```diff
@@ -27,6 +34,10 @@ allprojects {
             url "https://dl.bintray.com/touchlab/Squeaky"
         }
         maven { url "https://jitpack.io" }
+        maven {
```
liujoshua: Let's put google first, before jcenter. There was a recent issue (not sure if it's still happening) where a dependency was still being advertised in jcenter but was changed to only be available in google's artifact repo.

build.gradle
jdkizer9: Good idea, adding this

skin/build.gradle
jdkizer9: Fixing this

build.gradle
jdkizer9: Fixed this

build.gradle
jdkizer9: cleaned this up

PR #507, opened 2019-05-15
Feature/example app integration

jdkizer9:
Added example apps for foundation, backbone, and skin


skinsampleapp/src/main/java/org/researchstack/skinsampleapp/bridge/BridgeDataProvider.java
```diff
@@ -0,0 +1,1014 @@
...
+import retrofit2.http.POST;
+import retrofit2.http.Path;
+import rx.Observable;
+import rx.schedulers.Schedulers;
+
+/*
+* This is a very simple implementation that hits only part of the Sage Bridge REST API
+* a complete port of the Sage Bridge Java SDK for android: https://github.com/Sage-Bionetworks/BridgeJavaSDK
+ */
+public abstract class BridgeDataProvider extends DataProvider
```
syoung-smallwisdom, line 85: Is Sage being used here as an example of how you *could* connect to services?

I'm curious about this b/c we do have an SDK for talking to rest services on Android, but I don't think we want this repo to point to that, yeah?  

I kinda feel like an app that talks to services isn't a good choice for the base repo b/c those services change over time and it's kind of confusing to have a sample app that doesn't actually run.

syoung-smallwisdom (commented):
I have a concern about the inclusion of Bridge server access in the skin sample. Can that be refactored out and replaced with a "dummy" service or a debug print of the request/response?

skinsampleapp/src/main/java/org/researchstack/skinsampleapp/bridge/BridgeDataProvider.java
jdkizer9: Deprecated BridgeDataProvider and added comment that it should not be used. Note that Skin itself (and presumably Skin sample app) is deprecated.

PR #515, opened 2019-07-09
Feature/separate compabitility module

liujoshua:


jdkizer9:
So I took a look and I like how it makes it really easy for existing applications to migrate to 2.0. I think what I'm a little confused by is that it looks like at the moment I don't see how old Steps and new Steps can be used in the same Task. It looks like ViewBackboneInteropTaskActivity is doing some translation from old Step to new Step, but maybe we could make an InteropOrderedTask that supports both, probably via a StepAdaptorFactory with a different signature? Or maybe we could create an internal wrapper around the old Steps so that they acted like new Steps? Although this might provide some challenging result wrangling, but maybe that can be resolved with an InteropTaskResult.

Also, can we assume that the Task object that's passed in is an OrderedTask?

jdkizer9:
Ok, so I looked into it, and previously both the backbone and the foundation sample apps had a back button in their app bar. The activity should probably be updated to add a back button. Also, when the back button is pressed, the fragment should be initialized based on saved step result for that step.

